PR #548, opened 2024-04-18
Use metric instead of growthrate for vertical stability

bclyons12:


bclyons12:
@orso82 We need to resolve this per our discussion today

PR #557, opened 2024-05-02
Tutorial fix v2

avdeevag:
Something what I could not fix:
- Looks like something is wrong with a current initialization for FFP case, or maybe I just suppose to take another FFP_v1 case from different directory?
- I could not find **IMAS.save** and **IMAS.load**
- **ActorCXbuild** does not work for this FFP case with `rebuild_wall=true`
- **ActorBlanket** does not work


TimSlendebroek (approved):
These changes make sense, it wasn't that much in the end!
There is a lot more functionality since then but that will be for a new tutorial :)

PR #593, opened 2024-07-02
Adding in ITB to ini

TimSlendebroek:
@orso82  this is a temp fix as we had a discussion on how we actually want to do it per channel but this will suffice for now for @tomneiser @nanshi1177  to work together

example

```julia
ini,act= FUSE.case_parameters(:ITER;init_from=:scalars);

#ini.core_profiles.ITB_radius
ini.core_profiles.ITB_radius = 0.6
ini.core_profiles.ITB_width = 0.05
ini.core_profiles.ITB_height_temperature  = 2e3
ini.core_profiles.ITB_height_density = 0.3e20
```

image

### DIFF
--- a/src/ddinit/init_core_profiles.jl
+++ b/src/ddinit/init_core_profiles.jl
@@ -62,6 +62,10 @@ function init_core_profiles!(dd::IMAS.dd, ini::ParametersAllInits, act::Paramete
                 ini.core_profiles.ne_setting,
                 ini.core_profiles.ne_value,
                 ini.core_profiles.T_ratio,
+                ITB_radius=getproperty(ini.core_profiles, :ITB_radius, missing),
+                ITB_width=getproperty(ini.core_profiles, :ITB_width, missing),
+                ITB_height_temperature=getproperty(ini.core_profiles, :ITB_height_temperature, missing),
+                ITB_height_density=getproperty(ini.core_profiles, :ITB_height_density, missing),                
                 ini.core_profiles.T_shaping,
                 ini.core_profiles.n_shaping,
                 ini.core_profiles.ne_sep_to_ped_ratio,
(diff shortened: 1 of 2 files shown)


orso82:
@TimSlendebroek I'd like ITB to be it's own standalone function that can be applied to any profile. Like this:
```julia
function ITB(rho0, width, height, n)
    rho = range(0.0, 1.0, n)
    return @. (tanh((-rho + rho0) / width * pi / 2) + 1.0) / 2.0 * height
end

rho0 = 0.5
width = 0.25
height = 100.0
n = 101

h = ITB(rho0, width, height, n)
p = IMAS.Hmode_profiles(80.0, 500.0, n, 1.2, 1.2, 0.1)

plot(r, p, label="original")
plot!(r, p.+h, label="original + ITB")
plot!(r, h, label="ITB alone")
vline!([rho0-width, rho0+width], label="width", ls=:dash, color=:black)
```

image

After talking to @nanshi1177 we agreed `ini` should take in the following:
```julia
ini.core_profiles.ITB_radius
ini.core_profiles.ITB_ne_width
ini.core_profiles.ITB_ne_ratio
ini.core_profiles.ITB_T_width
ini.core_profiles.ITB_T_ratio
```
essentially allowing density and temperature ITBs to have different widths and heights. Also it would be best if the heights were expressed as a ratio of the value of the profile without ITB evaluated at `rho0`.

orso82:
@nanshi1177 @tomneiser @TimSlendebroek with this commit https://github.com/ProjectTorreyPines/FUSE.jl/commit/f8d44089cb4f0fe14a5642769eb9def3b2c0cdb3 now one can add an ITB to any profile at the `init()` stage. For example:

image

@fmunguia-wulf , after giving it some thought, instead of anding an artificial current source to mimic the effect of an ITB, I think it's better to just start your simulations with profiles that have an ITB. This will give you the right bootstrap current for those profiles. In the example above, like this:

image

This is on the `dev` branch, and I'll merge it once the regression tests pass

PR #704, opened 2024-09-10
try running FUSE on Windows

orso82:


bclyons12:
Post-APS, I'm going to spend some time trying to run exclusively on a Windows machine, so this is good.

PR #714, opened 2024-09-19
add data_flow_external

TimSlendebroek:
replaces use_actor_u


TimSlendebroek:
we figured this out in a better way that doesn't use external and put the complicated calls inside bop

PR #717, opened 2024-09-24
Use properly computed Ip_non_inductive in ActorQED

bclyons12:


orso82:
@bclyons12 see

https://github.com/ProjectTorreyPines/FUSE.jl/pull/721

https://github.com/ProjectTorreyPines/IMAS.jl/pull/203

PR #719, opened 2024-09-25
Plasma evolution with feedforward coils

bclyons12:
This is working towards a Grad-Hogan solver. To start, we're going to take the coil voltages and p' & FF' as a function of time from an existing TokSys simulation, then do a simple loop between FRESCO equilibrium solver and QED with build evolution.

13B4759C-0600-4C28-8099-B410ADCD5038


bclyons12:
From @anandhimank: location of the eqdsks '/fusion/projects/codes/toksys/users/anandh/ITER_data/EQDSK/14652.eqdsk'

bclyons12:
TokSys results from @anandhimank : `/fusion/projects/codes/toksys/users/anandh/toksyslearn/toksys/tokamaks/iter/sim/GSevolve/For_lyons_FPP_IRAD.mat`

PR #720, opened 2024-09-25
Using BalanceOfPlantSurogate to predict plant efficiency at lightspeed!

TimSlendebroek:


orso82:
@TimSlendebroek this is fantastic! Truly great work!

I'd like to bring in the BalanceOfPlantSurrogate model as part of this existing PR https://github.com/ProjectTorreyPines/FUSE.jl/pull/596 which makes ThermalSystemsModels an optional dependency in FUSE. This will relieve us of having the ModelingToolkit dependency, which as you know is very heavy!

Also, I have made few touchups to BalanceOfPlantSurrogate https://github.com/ProjectTorreyPines/BalanceOfPlantSurrogate.jl/pull/1 that we should include before moving forward!

PR #722, opened 2024-09-30
Solid mechanics development

daveweisberg:
### Summary:

The option to have a "nose" structure as part of the TF coil has been added. 

This is controlled by the parameter `nose_thickness`, which is set in `ini.tf.nose_thickness` and is stored in the data structure at `dd.build.tf.nose_thickness`. 

The TF nose is a solid steel structure on the inboard part of the HFS TF coil, and has a radial thickness equal to `nose_thickness` multiplied by the TF width, where `nose_thickness` is a float between 0.0 and 1.0. Increasing the size of the TF nose will decrease the available space for the TF conductor, however this trade-off is usually favorable and is a useful optimization parameter.

### Modified files:

- **stresses_actor.jl** has been modified to calculate the freestanding TF stresses with this new TF nose as a part of the mechanical system. NOTE: other coil configurations (bucked OH-TF, bucked plug-OH-TF, bucked plug-TF) have not yet been updated to allow for a TF nose. 
- **hfs_actor.jl** has been modified to allow for a TF nose, and the centerpost optimization has been updated to include the TF `nose_thickness` as an optimization parameter.
- **tf_magnet.jl** has been modified to use `nose_thickness` in calculating the cross-sectional area of the TF conductor.

### Unfinished work:

This change to the TF build will require modifications to the costing actors, so that they will correctly calculate the volume of SC conductor and steel support in the TF coils. I haven't done this yet, and I would appreciate any views on the best way to do this. 

Specifically, it seems that the costing functions only use the build layer information and the `build.tf.technology` information. However, I have placed the `nose_thickness` parameter directly in `build.tf`. Would it make more sense to put it in `build.tf.technology`?


orso82:
@daveweisberg I wonder if the nose of the TFs should just appear as a separate layer in the build.

daveweisberg:
@orso82 Yes that's certainly a possibility. On one hand, the TF nose has distinct material composition from the rest of the TF. On the other hand, the TF coil is usually thought of as a single component. Furthermore, the TF nose wouldn't be a meaning full stand-alone layer - it would need to exist in a specific relationship to the TF coil.

I think the way I have it in this commit makes the most sense, but I'm not too strongly attached. If you think another way is better, please go ahead.

orso82:
Uhm... let me think a bit more about the implications of the two approaches...

daveweisberg:
@orso82 What's holding up this PR? Let me know if you need to talk more about data structure options.

src/parameters/parameters_inits.jl
```diff
@@ -169,6 +169,7 @@ Base.@kwdef mutable struct FUSEparameters__tf{T} <: ParametersInit{T}
     ripple::Entry{T} =
         Entry{T}("-", "Fraction of toroidal field ripple evaluated at the outermost radius of the plasma chamber"; default=0.01, check=x -> @assert x > 0.0 "must be: ripple > 0.0")
     technology::Switch{Symbol} = Switch{Symbol}(FusionMaterials.supported_coil_techs(), "-", "TF coils technology")
+    nose_thickness::Entry{Float64} = Entry{Float64}("-", "Relative thickness of the TF nose, as a fraction of total TF radial build"; default=0.0)
```
orso82, line 172: Why expressing nose thickness as a function of the total radial build? It's very hard to guess a reasonable initial value!

It would be a lot more intuitive to express it as a fraction of the TF radial build on the high-field-side... say 10%.

Maybe better calling it `nose_hfs_fraction`?

src/actors/build/stresses_actor.jl
```diff
@@ -152,6 +165,7 @@ function solve_1D_solid_mechanics!(
     R_oh_out::T;                           # : (float) major radius of outboard edge of OH coil, meters
     axial_stress_tf_avg::T=NaN,            # : (float) average axial stress in TF coil core legs, Pa (if nothing, use constant fraction of hoop stress)
     axial_stress_oh_avg::T=NaN,            # : (float) average axial stress in OH coil, Pa (if nothing, use constant fraction of hoop stress)
+    tf_nose_thickness::T=0.0,              # : (float), thickness of solid nose section of TF coil, normalized to TF coil thickness
```
orso82, line 168:
```suggestion
    tf_nose_thickness::T,              # : (float), thickness of solid nose section of TF coil, normalized to TF coil thickness
```

src/actors/build/stresses_actor.jl
```diff
@@ -113,6 +115,7 @@ end
         R_oh_out::T;                           # : (float) major radius of outboard edge of OH coil, meters
         axial_stress_tf_avg::T=NaN,            # : (float) average axial stress in TF coil core legs, Pa (if nothing, use constant fraction of hoop stress)
         axial_stress_oh_avg::T=NaN,            # : (float) average axial stress in OH coil, Pa (if nothing, use constant fraction of hoop stress)
+        tf_nose_thickness::T=0.0,              # : (float), thickness of solid nose section of TF coil, normalized to TF coil thickness
```
orso82, line 118:
```suggestion
        tf_nose_thickness::T,              # : (float), thickness of solid nose section of TF coil, normalized to TF coil thickness
```

src/actors/build/stresses_actor.jl
```diff
@@ -74,6 +75,7 @@ function _step(actor::ActorStresses)
             oh_on ? Bz_oh : 0.0,
             R_oh_in,
             R_oh_out;
+            tf_nose_thickness=tf_nose_thickness,
```
orso82, line 78:
```suggestion
            tf_nose_thickness,
```

orso82 (changes requested):
@daveweisberg this is great, and I agree with how the nose is implemented as part of the TF itself. I think this works because the thickness of the TF on the high and low field sides can be different.

My main concern at this point is with how the nose thickness is specified. expressing nose thickness as a function of the total radial build? It's very hard to guess a reasonable initial value, and very easy to come up with unphysical solutions where the thickness of the nose is larger than the hfs TF thickness itself! It would be a lot more intuitive to express it as a fraction of the TF radial build on the high-field-side... say 10%.

src/parameters/parameters_inits.jl
daveweisberg: "Total TF radial build" here means exactly this - the thickness of the TF coil on the high-field side. We can the parameter description to more clearly communicate this. But the actual code is already doing what you suggest.

src/actors/build/stresses_actor.jl
daveweisberg: tf_nose_thickness = 0 means that there is no TF nose. Maintains default case.

orso82: yes, I got that. I would just remove the default value, essentially forcing the users of this function to explicitly take ownership of that parameter.

src/actors/build/tf_magnet.jl
```diff
@@ -10,7 +10,8 @@ NOTE: This function is here but it's not really used for machine desing, since g
 """
 function tf_maximum_J_B!(bd::IMAS.build; coil_j_margin::Float64)
     hfsTF = IMAS.get_build_layer(bd.layer, type=_tf_, fs=_hfs_)
-    TF_cx_area = π * (hfsTF.end_radius^2 - hfsTF.start_radius^2) / bd.tf.coils_n
+    conductor_start_radius = hfsTF.start_radius + bd.tf.nose_thickness*(hfsTF.end_radius-hfsTF.start_radius)
```
orso82, line 13:
```suggestion
    conductor_start_radius = hfsTF.start_radius + bd.tf.nose_thickness*hfsTF.thickness
```

src/actors/build/tf_magnet.jl
```diff
@@ -47,7 +48,8 @@ function tf_required_J_B!(bd::IMAS.build, eq::IMAS.equilibrium)
 
     # current in the TF coils
     current_TF = B0 * R0 * 2π / constants.μ_0 / bd.tf.coils_n
-    TF_cx_area = π * (hfsTF.end_radius^2 - hfsTF.start_radius^2) / bd.tf.coils_n
+    conductor_start_radius = hfsTF.start_radius + bd.tf.nose_thickness*(hfsTF.end_radius-hfsTF.start_radius)
```
orso82, line 51:
```suggestion
    conductor_start_radius = hfsTF.start_radius + bd.tf.nose_thickness*hfsTF.thickness
```

src/parameters/parameters_inits.jl
orso82: what threw me off was that you used `(hfsTF.end_radius-hfsTF.start_radius)` instead of `hfsTF.thickness`. This reinforces the point of using `nose_hfs_fraction`

src/actors/build/tf_magnet.jl
daveweisberg: I don't understand the difference between these two options. Is one prefered?

src/actors/build/stresses_actor.jl
orso82: It shouldn't. `solve_1D_solid_mechanics!` is only used in the `ActorStresses`. Any new run will have `nose_hfs_fraction` fraction defined.

Either way, this is just a minor stylistic comment, it's really not important.

src/actors/build/tf_magnet.jl
orso82: Functionally they are the same, but I rather use `.thickness` if you are referring to the thickness.

`bd.tf.nose_hfs_fraction * hfsTF.thickness` makes people understand right away what you are doing.

`bd.tf.nose_thickness*(hfsTF.end_radius-hfsTF.start_radius)` may be doing the same thing, but it's just harder to quickly grasp..

In general it's better if your code "reads" like what you are doing... if done right, this almost makes comments superfluous.

orso82: On a more technical level `.thickness` is a value in the tree, while `.end_radius` and `.start_radius` are expressions. This makes `.thickness` is much faster to evaluate.

daveweisberg:
@orso82 All suggested changes have been approved.

orso82:
@daveweisberg we can merge once the regression tests pass

orso82:
@daveweisberg there are some legit errors with the regression tests https://github.com/ProjectTorreyPines/FUSE.jl/actions/runs/11710913159/job/32618287199

I suspect the issue is with handling cases with bucking.

Please pull from latest FUSE, IMAS, IMASdd for the `solid_mech_dev` branch for testing

PR #742, opened 2024-10-28
Write neutron rate profile from RABBIT to dd

adrianaghiozzi:
See https://github.com/ProjectTorreyPines/RABBIT.jl/pull/3 in RABBIT

### DIFF
--- a/src/actors/hcd/nbi/rabbit_actor.jl
+++ b/src/actors/hcd/nbi/rabbit_actor.jl
@@ -93,6 +93,17 @@ function _finalize(actor::ActorRABBIT)
         ion.particles = source.profiles_1d[].electrons.particles
         ion.particles_inside = source.profiles_1d[].electrons.particles_inside
         ion.fast_particles_energy = beam_energy
+
+        # add neutron rate profile 
+        neutral = resize!(source.profiles_1d[].neutral, 1)[1]
+        neutral_particles = output.nrate_data[:, end]
+        neutral_particles = [neutral_particles[1]; neutral_particles; neutral_particles[end]]
+        neutral.particles = neutral_particles
+        neutral.label = "neutrons"
+
+        element = resize!(neutral.element, 1)[1]
+        element.z_n = 0.0
+        
     end
 
     return actor


src/actors/hcd/nbi/rabbit_actor.jl
```diff
@@ -93,6 +93,17 @@ function _finalize(actor::ActorRABBIT)
...
         ion.particles_inside = source.profiles_1d[].electrons.particles_inside
         ion.fast_particles_energy = beam_energy
+
+        # add neutron rate profile 
+        neutral = resize!(source.profiles_1d[].neutral, 1)[1]
+        neutral_particles = output.nrate_data[:, end]
+        neutral_particles = [neutral_particles[1]; neutral_particles; neutral_particles[end]]
+        neutral.particles = neutral_particles
+        neutral.label = "neutrons"
+
```
orso82, line 103: @adrianaghiozzi I was wondering if RABBIT distinguishes between neutrons of different energies...
```
D+T→He4   (3.518 MeV)  + n (14.072 MeV) + 17.59MeV

D+D→He3   (0.8175 MeV) + n (2.4525 MeV) + 3.27MeV
D+D→T     (1.0075 MeV) + H (3.0225 MeV) + 4.03MeV

D+He3→He4 (3.66 MeV)   + H (14.64 MeV) + 18.3MeV
```

The more I look at `core_sources.source[:].profiles_1d[].neutral` as a way to store neutrons, the more I come to the conclusion that we should extend `core_sources` with `core_sources.source[:].profiles_1d[].neutron`. How about with something like this?

```
 neutrons[:]
 ├─ label [] {String}
 ├─ index [] {Int} # 1: fusion, 2:beam-target, 3:beam-beam
 ├─ energy [eV] {Float}
 ├─ particles [s^-1.m^-3] Vector{Float}
 └─ particles_inside [s^-1] Vector{Float}  # mostly for symmetry with electrons and ions particle sources, for us it will be an expression
```

adrianaghiozzi: I think this is a good idea in general but we wouldn't be able to apply it to RABBIT without directly editing the RABBIT source to give more outputs. Right now RABBIT sums up the neutron rate contributions of thermonuclear, beam-target and beam-beam reactions and writes only the summed result to file. Similarly it considers the neutron energies separately as well based on the reaction type at some point, but seemingly sums away the individual contributions. 

Screenshot 2024-10-30 at 2 39 45 PM


See below comparing the output of RABBIT for the total neutron rate to the sum of the individual contributions which are output in this case by NUBEAM 
Screenshot 2024-10-30 at 2 37 29 PM

PR #747, opened 2024-11-05
add iter pf_passive from toksys

lstagner:
```julia
ini, act = FUSE.case_parameters(:ITER; init_from=:ods);
dd = IMAS.dd()
dd = FUSE.init(dd, ini, act);

plot(dd.equilibrium; color=:gray, cx=true)
plot!(dd.pf_active)
plot!(dd.pf_passive,color=:gray)
```
image


bclyons12:
So the passive structures look like this, and there are 148 of them. That's all consistent with the TokSys model?
Screenshot 2024-11-05 at 3 16 39 PM
If so, this is good to merge.

lstagner:
the ods was directly taken from the tok_data_struct that I have been using. The resistance was also taken from there with the resistivity calculated via the new resistitivity method in VacuumFields.

orso82:
Just a comment that I dislike these quads that overlap, but I understand that's how things are done in EFIT and elsewhere.

orso82:
@lstagner now we have 148 loops of one element each. I suspect we'd want to have four loops, with multiple elements each, essentially grouping them by `passive inner plate`, `passive outer plate`, `inner vessel wall`, `outer vessel wall`.

bclyons12:
@orso82 I think @lstagner  has it correct. The loops are in parallel, not series, so they shouldn't be connected. The pf_passive IDS requires that each loop has a current, not each element, so we want them as all separate loops.

orso82:
great! that solves it!

PR #751, opened 2024-11-09
Improve robustness of ActorVerticalStability

bclyons12:


bclyons12:
@orso82 note the error in D3D's pf_passive. And then there is more work we could do here, but this should all work once we merge `dev` into VacuumFields `master`. 

Eventually, I think a full refactoring of VacuumFields would be appropriate, but another time.

PR #763, opened 2024-11-19
Update stationary_plasma_actor.jl

daveweisberg:
Add pedestal actor in Stationary_Plasma actor for EPEDProfiles transport options.

### DIFF
--- a/src/actors/compound/stationary_plasma_actor.jl
+++ b/src/actors/compound/stationary_plasma_actor.jl
@@ -61,6 +61,11 @@ function ActorStationaryPlasma(dd::IMAS.dd, par::FUSEparameters__ActorStationary
             zeff_ped_from=:pulse_schedule,
             rho_nml=actor_tr.tr_actor.par.rho_transport[end-1],
             rho_ped=actor_tr.tr_actor.par.rho_transport[end])
+    elseif act.ActorCoreTransport.model == :EPEDProfiles
+        actor_ped = ActorPedestal(
+            dd,
+            act.ActorPedestal,
+            act)
     else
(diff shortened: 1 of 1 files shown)


orso82:
@daveweisberg why not simply doing
```
if act.ActorCoreTransport.model in (:FluxMatcher, :EPEDProfiles)
```

daveweisberg:
@orso82 I want the user to have control over what PedestalActor settings to use for EPEDProfiles. Hence the pass through of act.ActorPedestal.

orso82:
what settings?

orso82:
I'd really want to avoid this a different behavior depending on the `actor`. What if we did the following?

```julia
if ismissing(act.ActorPedestal, :rho_nml)
    rho_nml = actor_tr.tr_actor.par.rho_transport[end-1] 
else
    @warn "Taking `act.ActorPedestal.rho_nml` from user rather than from `act.ActorTransport.rho_transport` grid"
    rho_nml = act.ActorPedestal.rho_nml
end
```

and the same for `rho_ped`

Then we can use `if act.ActorCoreTransport.model in (:FluxMatcher, :EPEDProfiles)`

daveweisberg:
Here's my thinking. The existing code has a specific combination of ActorPedestal settings that are used for :FluxMatcher, and I'm assuming that these were chosen for specific reasons related to the FluxMatcher actor. 

However, the EPEDProfiles actor doesn't share the same requirements, and so would allow for a wider range of Pedestal settings. 

So for EPEDProfiles can we let all the Pedestal settings pass through? This will allow for more flexibility on my end for testing and experimentation.

PR #764, opened 2024-11-19
Major update to the multi_objective and database generator studies

TimSlendebroek:
Works fairly smoothly now :)

@nanshi1177
nanshi1177  you can pull this if you want to have the new way of saving files!


TimSlendebroek:
@orso82 I am thinking of adding a small little test for the database generator so that the study stuff is at least somewhat tested, what do you think?

orso82:
You can add the test, but make it conditional to `if get(ENV, "FUSE_WITH_EXTENSIONS", "false") == "true"` which is the case for the tests that we run once a day. We may change the name of the environmental variable later to something like `DETAILED_TESTS` or something like this.

orso82:
Also, keep the test as minimal as reasonably possible.

TimSlendebroek:
makes sense , something like the latest commit @orso82 ?

orso82:
yes, great! without the cruft (and formatting)

studies/MultiObjectiveOptimizer.jl
```diff
@@ -27,6 +27,7 @@ Base.@kwdef mutable struct FUSEparameters__ParametersStudyMultiObjectiveOptimize
     n_workers::Entry{Int} = study_common_parameters(; n_workers=missing)
     file_save_mode::Switch{Symbol} = study_common_parameters(; file_save_mode=:safe_write)
     release_workers_after_run::Entry{Bool} = study_common_parameters(; release_workers_after_run=true)
+    run_in_safe_mode :: Entry{Bool} = Entry{Bool}("-", "Runs the optimization in a loop restarting the workers every 5 generations"; default=false)
```
orso82, line 30: `run_in_safe_mode` is not a good name. Maybe `restart_workers_after_n_generations` ? also, this should be an `Int`, not a `Bool` with the ability to set after how many generations the workers would be restarted. Default should be `0` to mean no restart.

orso82 (commented):
@TimSlendebroek looks great. I have some comments, mainly on how restart workers after N generations should work.

Also, please apply formatting throughout.

Finally, has the FuseExample notebook been updated?

PR #766, opened 2024-11-20
Build bug fix

daveweisberg:
Minor fixes to build actors, including: rectangle_ellipse and top/bottom clearance calculation. 

Also updated FPP case file.


src/actors/build/cx_actor.jl
```diff
@@ -886,7 +886,11 @@ function optimize_layer_outline(
         func = shape_function(shape; resolution)
         initial_clerance = max(hfs_thickness, lfs_thickness) * vertical_clearance
         shape_parameters0 = initialize_shape_parameters(shape, oR, oZ, l_start, l_end, initial_clerance)
-        shape_parameters = optimize_outline(oR, oZ, hfs_thickness, lfs_thickness, func, l_start, l_end, shape_parameters0)
+        if occursin("TF",layer.name)
+            shape_parameters = shape_parameters0
+        else
+            shape_parameters = optimize_outline(oR, oZ, hfs_thickness, lfs_thickness, func, l_start, l_end, shape_parameters0)
+        end
```
orso82, line 893: I also don't agree with this change. How can you know the right parameters for the TF layer if you don't optimize them?

src/actors/build/cx_actor.jl
```diff
@@ -961,7 +965,7 @@ function optimize_outline(
             if hfs_thickness != lfs_thickness
                 hbuf = hfs_thickness - target_clearance
                 lbuf = lfs_thickness - target_clearance
-                r_obstruction, z_obstruction = buffer(r_obstruction, z_obstruction, hbuf, lbuf)
+                r_obstruction, z_obstruction = buffer(r_obstruction, z_obstruction, hbuf, hbuf)
```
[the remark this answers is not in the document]
daveweisberg: The goal here is to have the up/down clearance be equal to the hfs gap, instead of the average of the hfs and lfs gaps. This way works for me, but I'm sure there's a better way to do it...

src/actors/build/cx_actor.jl
daveweisberg: I added this change because the princeton_D_scaled TF shape was being incorrectly changed by this line. I'm my experience, it makes more sense to have the TF shape be dictated by the user instead of optimized... But to be fair I've only been working with one specific case, and I don't know if this is generalizable.

orso82: The optimizer will try to make the layers (including the TF) be as snug as possible. Do you have an example of it failing with `princeton_D_scaled`? what kind of failure were you seing?

daveweisberg: Before my fix:
Screen Shot 2024-11-20 at 11 00 05 AM

After my fix:
Screen Shot 2024-11-20 at 11 02 18 AM

src/actors/build/cx_actor.jl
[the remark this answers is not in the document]
daveweisberg: Before my fix:
Screen Shot 2024-11-20 at 11 05 28 AM


After my fix:
Screen Shot 2024-11-20 at 11 02 18 AM

orso82:
The tighter way to generate a build and updated FPP have all made it now to master

orso82:
@daveweisberg please note that after I merged your changes to the FPP use case the regression tests were failing:
https://github.com/ProjectTorreyPines/FUSE.jl/actions/runs/12174530001/job/33956679939#step:17:1179
so I reverted back to the original FPP definition on `master`.

Please open a separate PR with the updates to the FPP case, and I'll make sure that the regression tests pass before merging into `master`

PR #769, opened 2024-11-20
Fix number of MXH coeffs in tequila2imas

### LINKED ISSUE #768: A mismatch between `grid.dim2` and the size of `psi` in `tequila2imas` function
# Issue in `profiles_2d[1]` (flux coordinate)

## A Mismatch Between `grid.dim2` and the Size of `psi` 

In the following example, `dim2` is set to 21, but the second dimension of `psi` (`size(psi, 2)`) is 13.

Screenshot 2024-11-20 at 11 43 24 AM

# Cause of the Issue

The following code block in the `tequila2imas` function defines `dim2`:

https://github.com/ProjectTorreyPines/FUSE.jl/blob/605a7ef0e98e76b81b986ec9d59986b6fa1a6716/src/actors/equilibrium/tequila_actor.jl#L187-L192

In this code block, `shot.M` (which corresponds to `number_of_fourier_modes` (default=8) in the `ActorTEQUILA` parameters) is used to define `dim2`. 

However, the size of `psi` is determined by `number_of_MXH_harmonics` (default=4), resulting in 13 components in the second dimension. (where 13 components = 5 basic terms + 4 cos terms + 4 sin terms).

# Questions

- What is the difference between `number_of_fourier_modes` and `number_of_MXH_harmonics` in the `FUSEparameters__ActorTEQUILA`? Do they have different applications?
- What are the correct definitions for `dim2` and the second dimension of `psi`?

bclyons12:
Fixes #768

@mgyoo86 please test

### DIFF
--- a/src/actors/equilibrium/tequila_actor.jl
+++ b/src/actors/equilibrium/tequila_actor.jl
@@ -188,7 +188,7 @@ function tequila2imas(shot::TEQUILA.Shot, dd::IMAS.dd{D}, par::FUSEparameters__A
     eq2d = eqt.profiles_2d[1]
     eq2d.grid.dim1 = shot.ρ
     MXH_modes = (size(shot.surfaces, 1) - 5) ÷ 2
-    eq2d.grid.dim2 = vcat([0.0, 0.0, 0.0, 0.0, 0.0], fill(1.0, (MXH_modes,)), fill(-1.0, (MXH_modes,)))
+    eq2d.grid.dim2 = vcat(@SVector[0.0, 0.0, 0.0, 0.0, 0.0], range(1,MXH_modes), .-range(1,MXH_modes))
     eq2d.grid_type.index = 57 # inverse_rhotor_mxh : Flux surface type with radial label sqrt[Phi/pi/B0] (dim1), Phi being toroidal flux, and MXH coefficients R0, Z0, ϵ, κ, c0, c[...], s[...] (dim2)
     eq2d.psi = collect(shot.surfaces')
 
--- a/src/actors/equilibrium/tequila_actor.jl
+++ b/src/actors/equilibrium/tequila_actor.jl
@@ -187,7 +187,8 @@ function tequila2imas(shot::TEQUILA.Shot, dd::IMAS.dd{D}, par::FUSEparameters__A
     # MXH flux surface parametrization
     eq2d = eqt.profiles_2d[1]
     eq2d.grid.dim1 = shot.ρ
-    eq2d.grid.dim2 = vcat([0.0, 0.0, 0.0, 0.0, 0.0], fill(1.0, (shot.M,)), fill(-1.0, (shot.M,)))
+    L = length(surfaces[1].c)
+    eq2d.grid.dim2 = vcat([0.0, 0.0, 0.0, 0.0, 0.0], fill(1.0, (L,)), fill(-1.0, (L,)))
     eq2d.grid_type.index = 57 # inverse_rhotor_mxh : Flux surface type with radial label sqrt[Phi/pi/B0] (dim1), Phi being toroidal flux, and MXH coefficients R0, Z0, ϵ, κ, c0, c[...], s[...] (dim2)
     eq2d.psi = collect(shot.surfaces')
 
--- a/src/actors/equilibrium/tequila_actor.jl
+++ b/src/actors/equilibrium/tequila_actor.jl
@@ -187,8 +187,8 @@ function tequila2imas(shot::TEQUILA.Shot, dd::IMAS.dd{D}, par::FUSEparameters__A
     # MXH flux surface parametrization
     eq2d = eqt.profiles_2d[1]
     eq2d.grid.dim1 = shot.ρ
-    L = length(shot.surfaces[1].c)
-    eq2d.grid.dim2 = vcat([0.0, 0.0, 0.0, 0.0, 0.0], fill(1.0, (L,)), fill(-1.0, (L,)))
+    MXH_modes = (size(shot.surfaces, 1) - 5) ÷ 2
+    eq2d.grid.dim2 = vcat([0.0, 0.0, 0.0, 0.0, 0.0], fill(1.0, (MXH_modes,)), fill(-1.0, (MXH_modes,)))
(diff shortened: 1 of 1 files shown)


src/actors/equilibrium/tequila_actor.jl
```diff
@@ -187,7 +187,8 @@ function tequila2imas(shot::TEQUILA.Shot, dd::IMAS.dd{D}, par::FUSEparameters__A
     # MXH flux surface parametrization
     eq2d = eqt.profiles_2d[1]
     eq2d.grid.dim1 = shot.ρ
-    eq2d.grid.dim2 = vcat([0.0, 0.0, 0.0, 0.0, 0.0], fill(1.0, (shot.M,)), fill(-1.0, (shot.M,)))
+    L = length(shot.surfaces[1].c)
+    eq2d.grid.dim2 = vcat([0.0, 0.0, 0.0, 0.0, 0.0], fill(1.0, (L,)), fill(-1.0, (L,)))
```
mgyoo86, line 191: I've tested the current change, but it resulted in the following error

```
type Float64 has no field c
Stacktrace:
  [1] getproperty
    @ ./Base.jl:49 [inlined]
  [2] tequila2imas(shot::TEQUILA.Shot{Int64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, Matrix{Float64}, Matrix{Float64}, TEQUILA.Profile{FiniteElementHermite.FE_rep{Vector{Float64}, Vector{Float64}}, FiniteElementHermite.FE_rep{Vector{Float64}, Vector{Float64}}}, Nothing, Nothing, Nothing, TEQUILA.Profile{FiniteElementHermite.FE_rep{Vector{Float64}, Vector{Float64}}, FiniteElementHermite.FE_rep{Vector{Float64}, Vector{Float64}}}, Float64, Float64, Float64, FiniteElementHermite.FE_rep{StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, Vector{Float64}}, Vector{FiniteElementHermite.FE_rep{StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, Vector{Float64}}}, TEQUILA.QuadInfo{Vector{Float64}, Vector{SparseArrays.SparseVector{Float64, Int64}}, Matrix{Float64}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, Matrix{Float64}, Matrix{Float64}}, Vector{PreallocationTools.DiffCache{Vector{Float64}, Vector{Float64}}}, BandedMatrices.BandedLU{Float64, BandedMatrices.BandedMatrix{Float64, Matrix{Float64}, Base.OneTo{Int64}}}}, dd::IMASdd.dd{Float64}, par::FUSE.FUSEparameters__ActorTEQUILA{Float64}; ψbound::Float64, green_model::Symbol)
    @ FUSE ~/.julia/dev/FUSE/src/actors/equilibrium/tequila_actor.jl:190
  [3] _finalize(actor::FUSE.ActorTEQUILA{Float64, Float64})
    @ FUSE ~/.julia/dev/FUSE/src/actors/equilibrium/tequila_actor.jl:140
...
...
```

I've tried the following thing as you might intend, and it works fine.
Could you please check this? 
Thank you very much.

```suggestion
    N_MXH = length(IMAS.MXH(shot.surfaces[:, end]).c)
    eq2d.grid.dim2 = vcat([0.0, 0.0, 0.0, 0.0, 0.0], fill(1.0, (N_MXH,)), fill(-1.0, (N_MXH,)))
```

mgyoo86 (changes requested):

bclyons12:
@orso82 See, this is what happens if I just start pushing changes and don't test them first...

@mgyoo86 Sorry, try again. And by the way, anytime you do slicing, you want to use `@view` or `@views` to avoid allocations. For example:
```
A = rand(3,3)
B = zeros(3)
B .= A[:, 2] # allocates new vector of length 3, then broadcasts it
B .= @view(A[:,2]) # does not allocate, just a view of the original matrix
@views B .= A[:, 3] .+ A[1, :] # again, no allocations
```

mgyoo86:
@bclyons12 Thank you for the quick fix, it now works!
Also, I appreciate the great tips on using `@view` :)
Julia's allocations under the hood are still too magical thing for me  😭

BTW, this might be one of good use cases of the forked repo, where the GitHub actions run tests for my pushed commits on remote (not harshing my laptop), after which I can open the PR safely.

mgyoo86:
@bclyons12  
Before closing this PR, I have a question about the `dim2` for the flux-coordinate (`profiles_2d[1]`)
What is the meaning of the following definitnion of `dim2`?
```
eq2d.grid.dim2 = vcat([0.0, 0.0, 0.0, 0.0, 0.0], fill(1.0, (MXH_modes,)), fill(-1.0, (MXH_modes,)))
```
If there is no specific meaning, can we just set `dim2` as the index, like the following?
```
eq2d.grid.dim2 = collect(Float64, 1:length(MXH_modes))
```

The reason why I ask this follows.
In the following PR for the `IMAS` repo, I implemented new plot recipes, of which 3D plot recipe (contour, surface, ...) uses `dim1` and `dim2` as `x` and `y` axes. 

https://github.com/ProjectTorreyPines/IMAS.jl/pull/216

The problem is that I need `x` and `y` data sorted in ascending order to use recipes of `contour` or `heatmap`.
The new recipes work well with the (R,Z) coordinate, but will not work for the flux-coordinate as the `dim2` is not defined in sorted ascending order.

bclyons12:
In this case, `eq2d.psi` is storing not the flux, but the MXH coefficients of each flux surface. `eq2d.grid.dim1` is the normalized rho_toroidal, labeling each flux surface. `eq2d.grid.dim2` is supposed to represent the "mode type". The first 5 values are (R0, Z0, epsilon, kappa, c0). Those get "mode type" 0. Then the next `MXH_mode` values correspond to cosine terms and all get positive 1. The last `MXH_mode` values correspond to sine terms and get negative 1.

@orso, perhaps we should do
`eq2d.grid.dim2 = vcat([0.0, 0.0, 0.0, 0.0, 0.0], 1:MXH_modes, .-(1:MXH_modes))`?

orso82:
@bclyons12 I like your proposal! Since it carries more information and possibly allows skipping of modes.

PR #773, opened 2024-12-03
Flattop duration fix

daveweisberg:
Fixed bugs in `flattop_duration` matching, namely to remove the dependence on `coil_j_margin`. 

Previously, the output `flattop_duration` was forced be equal to the `requirements.flattop_duration * (1 + requirements.coil_j_margin)` 

Now the output `flattop_duration` is exactly equal to requirements.flattop_duration.

- Calculation of `flattop_flux_estimates` should not be scaled by `1 + coil_j_margin`
- `c_flt` should have a margin of 0.0, not `coil_j_margin`
- the margin for `flattop_duration` should not be dependent on `coil_j_margin`.

### DIFF
--- a/src/actors/build/fluxswing_actor.jl
+++ b/src/actors/build/fluxswing_actor.jl
@@ -64,7 +64,7 @@ function _step(actor::ActorFluxSwing)
     bd.flux_swing.rampup = rampup_flux_estimates(eqt, cp)
     bd.flux_swing.pf = pf_flux_estimates(eqt)
     if !ismissing(requirements, :flattop_duration) && !par.operate_oh_at_j_crit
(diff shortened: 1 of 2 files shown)


orso82:
@daveweisberg regression tests are failing.

Also, I am not sure I agree with the change. If one wants to have say 20% tolerance on the current limit on the ohmic solenoid, I believe that's equivalent to say that they want to have 20% tolerance on their flattop duration. In other words, the nominal flattop duration is the one cited at 20% below the absolute maximum, no?

daveweisberg:
@orso82 Ok, re-running tests with fix.

This change is important, because the previous way of doing this didn't match the main use case. What I want is a FUSE solution that has the requested `flattop_duration` when the CS coil current density is limited to the `coil_j_margin`. In practice, the CS coil would never be operated higher than this limit. So the actual flattop is never going to be longer.

PR #779, opened 2024-12-10
Fix pf plot.

NathanCummings:
Small fix in pf plotting that will prevent errors in (e.g.) FUSE.digest().


orso82:
Thank you for the commit @NathanCummings ! 🎉

PR #785, opened 2024-12-13
model to keep q>1 in stationary transport

jmcclena:
This pull request implements Jardin's model for stationary sawteeth to raise q>1. 

Example run 

```
using Revise
using FUSE
ini, act = FUSE.case_parameters(:D3D)


# Initialize the `dd` structure with 0D parameters
dd = FUSE.init(ini, act);
dd_init = deepcopy(dd)
#act.ActorStationaryPlasma.convergence_error = 0.05#1e-8
act.ActorStationaryPlasma.max_iter = 5
FUSE.ActorStationaryPlasma(dd,act)
```

Screenshot 2024-12-12 at 3 40 32 PM


jmcclena:
@bclyons12 Do you know if eta needs to be updated in the dd to get the correct flux consumption?

bclyons12:
I do not. I'd just have to look at how and wear the flux consumption is calculated.

I'll look this over next week.

orso82:
@jmcclena @bclyons12 looking more into this, what should happen to the non-inductive current inside of the inversion radius? I wonder if that non-inductive current get redistributed inside the inversion radius, perhaps using the same `flatten()` function that @jmcclena proposed?

bclyons12:
I know that bootstrap current relaxes on the electron collision time, though I'm not sure what happens if there's large MHD that's redistributing particles anyway. Nor do I know how other noninductive sources should behave.

Issue #786, opened 2024-12-13
Development ideas from code camp 


adrianaghiozzi:
**Core plasma physics** 
- Q profile optimization actor (rather than setting up a full multi-objective optimization run to tailor q profile)

**Edge plasma physics** 
- Enrichment: how many times more concentration you have of an impurity at the edge compared to in the core => Simple model that predicts the scaling of the impurity concentration profile: https://dx.doi.org/10.1088/1741-4326/ad3139 

**Engineering/build** 
 - Design of passive structures based on vertical stability => Rail or other defined area on which passive structures could be placed 
 - Calculation of the stresses on the PF coils => informs the amount of steel that goes into PF coils 

**Multi-objective optimization**
- Check function that ensures the user has set the corresponding requirements in ini for each constraint function selected 
- Simple functions for generating 2D histograms to analyze optimization results 
- Explain in some tutorial how to temporarily add a cluster for optimization runs rather than directly editing FUSE.parallel_environment source  
- Distance from the selected constraints, whether overshooting or undershooting, as a parameter in the risk => Use overall risk as an objective and close in on the constraints by penalizing in risk when you’re far away from constraints 
- Neural net based on optimization results, identifying trends => useful for non-linear uncertainty propagation 

**Backend/user interface** 
- Allow users to change materials costs by directly interfacing with FusionMaterials Material objects rather than changing FusionMaterials source 
- Show the expression that is being used to evaluate a certain field in dd 
- Bug: help_plot(dd.balance_of_plant) throws an error with :linewidth attribute (key :linewidth not found)
- Define a distribution structure that gets passed to ini and use that distribution to assign weights to the range of initialized parameters => Applicable to the database generator 
- Add a plot of plasma heat flux to digest 

**Time dependence**
- Add helper functions for setting up more complex ramp up phases => e.g. Go from circular bore to temporary lower single null to double null 
- Setting of density in ini in time dependent simulations - should the user be allowed to set a Greenwald density such that changes in Ip will update core density or should the user choose only pedestal density and allow core density to be fully determined by pellet source? 
- Allow resetting of PF coil optimizations based on equilibria for multiple time slices

mgyoo86:
@adrianaghiozzi 
The following PR is a fix for the plotting bug described in the `Backend/user interface` part.

https://github.com/ProjectTorreyPines/IMAS.jl/pull/223

adrianaghiozzi:
@mgyoo86 Thanks Min-gu!

PR #792, opened 2024-12-17
Add check function for ini.requirements vs constraint functions in optimization

adrianaghiozzi:
In the [code camp](https://github.com/ProjectTorreyPines/FUSE.jl/issues/786) there was a suggestion to add a check function in the multi-objective optimization setup that ensures that each requirement that is set in ini.requirements also has a constraint function that goes with it. This (+ a supporting PR in IMAS) implements that functionality, giving the user a warning if they set a requirement without setting a constraint function for it. 

Doing this also made me realize that we have no constraint function for TBR - is that an oversight or is there some particular reason? @TimSlendebroek

### DIFF
--- a/studies/MultiObjectiveOptimizer.jl
+++ b/studies/MultiObjectiveOptimizer.jl
@@ -67,6 +67,42 @@ function _setup(study::StudyMultiObjectiveOptimizer)
 
     parallel_environment(sty.server, sty.n_workers)
 
+    function check_matching(requirement::Symbol, constraint_functions::Vector{Symbol})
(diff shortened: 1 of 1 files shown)


TimSlendebroek:
@adrianaghiozzi this is a great addition, as for the TBR I never added it to the requirements as it always be able to get there due to the Li6/Li7 ratio it can change internally. It probably is good to make it an explicit requirement (and maybe we should think if a constraint to the Li7/Li6 ratio perhaps)

orso82:
I could see having constraint for a desired Li6/Li7 value

PR #804, opened 2025-01-24
Install files for FUSE sysimage and module on omega

bclyons12:
Putting the omega sysimage install scripts into version control, and updated to produce the desired module file.

The goal should be for the automated bot on Omega to sparse clone this directory and use the scripts like
`julia deploy/omega/install_fuse_environment.jl`


orso82:
@bclyons12 could you please update the OMEGA documentation as part of this PR?

bclyons12:
@orso82 Omega documentation is updated for users and developers. Note that `fuse` module doesn't work with the `gacode` module, since they each load different environments. I didn't touch that documentation. We could note that issue, or we could recompile gacode with the more modern gcc11.x module so it's compatible.

orso82:
Thank you @bclyons12 !!

Perhaps _recompile gacode with the more modern gcc11.x module_ on Omega is something that CSS could help us with

Ping @smithsp @jcandy

jcandy:
This is something we should discuss. I think it is important for FUSE to "coexist peacefully" with other software, specifically GACODE. I am happy to do what I can to make this happen. One downfall of OMFIT is that it over-engineers/hacks the software environment. I think FUSE should use something like numpy or scipy or DASH as models of peaceful coexistence. All these tools coexist peacefully with the entire software stack via pip. Is the idea that FUSE will install as a Julia package?

I also **do not** endorse the current GACODE module approach. Is FUSE going to be non-portable like OMFIT?

jcandy:
Also, isn't gcc/13.x the default on OMEGA?

orso82:
Thank you for your comments @jcandy .

Please note that [FUSE is already a Julia package](https://fuse.help/dev/install.html#FUSE-installation), and we rely on the (in my opinion excellent) Julia package manager to handle all of our dependencies. This is really working well for us, removing a lot of the overhead that I have experienced working with large Python projects.

We've installed FUSE on Macs (intel and Apple silicon), Linux (x86 and ARM), and even Windows (x86 both native and with WSL). Again, I attribute the portability mainly to Julia's packaging system. It just works. 💪 Let us know if you encounter any issue installing it!

orso82:
The documentation about running FUSE on Omega is now live: https://fuse.help/dev/install_omega.html
Thank you @bclyons12 !

PR #817, opened 2025-02-19
WIP: Update deploy scripts, including new one-liner

bclyons12:


orso82:
@bclyons12 we should also add the `FuseRegistry` to the registries supported by the public build

bclyons12:
@orso82 Yes, the FuseRegistry is already available. For example, I can add HelpPlots to a Julia environment when I've done `module load fuse`.

If you've done `module load julia`, every user needs to add the registry themselves.

bclyons12:
This was used to make the latest v0.6.0 installation. It can be installed by just running the deploy.sh script. There is now a `fuse` command to run with the sysimage in a read-only environment.

PR #839, opened 2025-03-17
User can define active_x_points in FRESCO

bclyons12:
Makes use of new active_x_points capability in FRESCO. Will wait to merge until the control-point definition can be refactored and unified between FUSE, FRESCO, and VacuumFields.


bclyons12:
@chrismarsden7 this should work. You can either checkout the `active_x_points` branch on both FRESCO and FUSE if you're in `dev` mode, or do `] add FUSE#active_x_points FRESCO#active_x_points`. Then you'll set
`act.ActorFRESCO.active_x_points = [1,2]` and it should work. For example:

```
using FUSE
dd, _, act = FUSE.load("fuse_data_100325/pre_coil_changes"; load_ini=false)
act.ActorPFactive.strike_points_weight=1.0
act.ActorPFactive.x_points_weight = 3.0
act.ActorEquilibrium.model = :FRESCO
act.ActorFRESCO.active_x_points=[1,2]
act.ActorFRESCO.relax=0.25

FUSE.ActorStationaryPlasma(dd, act)

plot(dd.equilibrium)
```

Screenshot 2025-03-17 at 1 52 43 PM

PR #850, opened 2025-03-26
missing logic when turbulence model is not TGLF

TimSlendebroek:
this shows how little we use other turbulence models


bclyons12:
Merging, but will wait until tests pass on master before releasing.

PR #852, opened 2025-03-28
Use FRESCO profiles fixed to rho_tor_norm by default

bclyons12:
Optional let's users go back to a psi_norm grid. Also accounts for FRESCO API changes.


bclyons12:
FUSE and FRESCO master are incompatible until this merge. Once the tests pass, I'll merge this in.

PR #865, opened 2025-04-03
save both TGLF and TJLF input files for TJLF model

avdeevag:


TimSlendebroek:
@avdeevag looks like orso recently pushed some changes and took the save_tglf stuff out of the fluxmatcher

another thing is that now that you have your wonderful save(inputTJLF) all you have to do is 

save(input_tglf) (note that this will now work for both InputTGLF and InputTJLF!

and no more need to convert files

PR #866, opened 2025-04-04
Save tjlf input files

TimSlendebroek:


src/actors/transport/tglf_actor.jl
```diff
@@ -253,3 +258,18 @@ function Base.show(io::IO, ::MIME"text/plain", input::Union{InputTGLF,InputTJLF}
...
+"""
+    save(input::Union{TGLFNN.InputCGYRO, TGLFNN.InputQLGYRO,  TGLFNN.InputTGLF, }, filename::String)
+
+Common save method for all the various inputTGLF types
+"""
+function save(input::Union{TGLFNN.InputCGYRO, TGLFNN.InputQLGYRO,  TGLFNN.InputTGLF, }, filename::String)
+    if input isa TGLFNN.InputCGYRO || input isa TGLFNN.InputQLGYRO || input isa TGLFNN.InputTGLF
+        return TGLFNN.save(input, filename)
+    elseif input isa TJLF.InputTJLF
+        return TJLF.save(input, filename)
```
avdeevag, line 271: For debugging TJLF code it would be very useful to save both TGLF (TGLFNN.save) and TJLF inputs (TJLF.save)

TimSlendebroek: good idea I will add a debug bool for that

TimSlendebroek:
(still need to implement this in the ActorStationaryPlasma) [reminder for self]

orso82:
Thank you @TimSlendebroek !

TimSlendebroek:
@orso82 when thinking about how I would implement saving input_tjlfs during the run

Adding in this to ActorStationaryPlasma feels icky
    save_input_tglfs_to_folder::Entry{String} = Entry{String}("-", "Save the intput.tglf files in designated folder"; default="")

This will over time pollute act.ActorStationaryPlasma a lot

I was thinking of what if we add some type of measurement actor that can be configured however you want and then add that actor run at a particular point in the code.

```julia
            # run equilibrium actor with the updated beta
            ProgressMeter.next!(prog; showvalues=progress_ActorStationaryPlasma(total_error, actor, actor.actor_eq))
            finalize(step(actor.actor_eq))

            # measurement actor
            ProgressMeter.next!(prog; showvalues=progress_ActorStationaryPlasma(total_error, actor, actor.actor_measure))
            finalize(step(actor.actor_measure))
```

orso82:
I am not sure `measurement` is a good name here.

A general approach could be to add a callback in the `step()` and `finalize()` calls. Like this, conceptually:

```julia
function step(actor::Actor)
    _pre_step(actor)
    _step(actor)
    _post_step(actor)
end

function finalize(actor::Actor)
    _pre_finalize(actor)
    _finalize(actor)
    _post_finalize(actor)
end
```

Then one can dispatch on each specific actor as they want (in a user notebook)
```
function _post_finalize(actor::ActorTGLF)
    save(actor, "filename")
end
```

TimSlendebroek:
Something like that would make sense but how would you control it inside the convergence loop inside the stationaryplasma actor for example

orso82:
@TimSlendebroek you can try this on the latest `dev` https://github.com/ProjectTorreyPines/FUSE.jl/commit/3fc4892983d54969ee793eb26617e2e00f0505af

TimSlendebroek:
I'll merge it for now can you message me when the callback stuff is in master @orso82 

also do you know why the DIII-D reg test fails?

PR #871, opened 2025-04-16
Give initial target Ip to QED

bclyons12:
This specifies a desired total plasma current, so the edge value of <∇ρ²/R²> within QED gets modified to match this exactly. 
To do:

- [ ] Figure out how/when the equilibrium grids in core_profiles should get updated. Updating the area to be consistent between core_profiles and equilibrium should make the discrepancy that needs to be corrected for smaller


bclyons12:
Some tests are failing. Any thoughts @orso82?

orso82:
things pass on OSX but fail on Linux. It's what we have now observed several times with Anderson solver in FluxMatcher. If you can reproduce it I have few more ideas to drill down why/where it happens

PR #879, opened 2025-04-24
Use VacuumFields refactor for current per turn

bclyons12:


bclyons12:
Tests passed at [f854c04](https://github.com/ProjectTorreyPines/FUSE.jl/pull/879/commits/f854c043c34d82b0604e7d3fec1fa53e8da99ce9) and the latest just updated some compats, which resolve properly. Merging.

PR #880, opened 2025-04-24
WIP: remove GS_IMAS_pf_active__coil

bclyons12:
The goal here is to use VacuumFields coils more broadly in ActorPFdesign, ideally getting rid of the GS_IMAS_pf_active__coil infrastructure which is kind of a Frankenstein of IMAS and VacuumFields.

Currently this is only partial work, which I'm saving for posterity.


orso82:
@bclyons12 what's the status of this?

PR #881, opened 2025-04-25
Flip sign of EC steering_angle_pol

bclyons12:
I'm not sure if this is "correct", but it fixes the error in the EXCITE test for me. Perhaps this just undoes the change in https://github.com/ProjectTorreyPines/IMAS.jl/pull/253 so it'll just cause problems elsewhere though?


bclyons12:
This was the wrong fix.

PR #883, opened 2025-04-28
Fix studies to use correct OverrideParameters

TimSlendebroek:


TimSlendebroek:
@orso82 this should fix wrapping in OverrideParameters, can you take a look so we can merge this asap (as the studies are broken)

orso82:
thanks for the prompt fix @TimSlendebroek !

PR #884, opened 2025-04-29
Use NonlinearSolve interface in ActorFluxMatcher

bclyons12:
JuliaHub recommended we use NonlinearSolve.jl instead of NLsolve.jl directly so they could more easily provide us support. That change is made here. Currently there is no way to pass `xtol` to NLsolve through NonlinearSolve, so I just raised the `ftol` parameter to prevent excessive iterations. We'll see how the tests do with that.

This issue is relevant to the ftol/xtol thing: https://github.com/SciML/NonlinearSolve.jl/issues/593


bclyons12:
@orso82 The DTT test is now failing for both architectures. We're still getting different convergence percentages between the two architectures though.

bclyons12:
@jmcclena This change opens up many more solvers for us to try: https://docs.sciml.ai/NonlinearSolve/stable/
There are both solvers native to the package and wrapped solvers (like the NLsolve solvers we're using now). I tried a few other solvers (including the default `FastShortcutNonlinearPolyalg`), but many of them diverge or are quite slow. Note that for any solver that would like to use autodiff for the Jacobian, the `flux_match_errors` function is not currently autodiffentiable, since we try to store things in arrays that assume `Float64` and ForwardDiff uses `Dual` types. I'm sure we could make it autodifferentiable, but for now you can tell it to use finite differences if the algorithms needs them. For example:
```
# This runs with finite differences, but diverges
alg = NonlinearSolve.FastShortcutNonlinearPolyalg(; autodiff = NonlinearSolve.ADTypes.AutoFiniteDiff())
```

bclyons12:
@TimSlendebroek @tomneiser may be interested as well

PR #890, opened 2025-05-01
Use NonlinearSolve Broyden method in FluxMatcher

### LINKED ISSUE #889: Use built-in NonlinearSolve solvers in ActorFluxMatcher
We know that we're getting some non-deterministic behavior that is architecture and compiler dependent (e.g., different results if `julia --check-bounds`) is used. This is, at least in large part, caused by the behavior of the NLsolve.jl `:anderson` solver we're using in ActorFluxMatcher. #884 converted to using the NLsolve solvers through the NonlinearSolve.jl package, but we see the same behavior since the underlying solver is the same. After talking to @bradcarman of JuliaHub, he suggested the following via email:

> Instead of using the NLsolveJL solver, can you try one of the Recommended Methods: [Nonlinear System Solvers · NonlinearSolve.jl](https://docs.sciml.ai/NonlinearSolve/stable/solvers/nonlinear_system_solvers/#Recommended-Methods)
> 
> Now that you've got NonlinearSolve.jl wired up, you simply need to change the algorithm (alg).  If you are experiencing issues with the NonlinearSolve built in solvers, then I can get Chris or Oscar from JuliaHub to offer some advice, they will know exactly which knobs to turn.  Using NLsolveJL we are not in control and can't offer much help.
> 
> The good news is by switching to the built in NonlinearSolve.jl solvers you should also benefit from a faster solve.

I tried a few of the suggested methods but they either diverged or were slow. It would be good if someone with more FluxMatcher experience could dig into this and try it out.

bclyons12:
This seems to be fast and to converge robustly in my testing. Since it's a NonlinearSolve native solver, hopefully this fixes the architecture dependent issue we have. Potentially closes #889 

- [ ]  Maybe make default value of algorithm solver dependent? @jmcclena says this might not work with full TJLF because it uses a (finite-difference) Jacobian
- [ ] Could switch to [FixedPointAcceleration](https://docs.sciml.ai/NonlinearSolve/stable/api/fixedpointacceleration/) for Anderson as well

### DIFF
--- a/src/actors/transport/flux_matcher_actor.jl
+++ b/src/actors/transport/flux_matcher_actor.jl
@@ -25,7 +25,7 @@ Base.@kwdef mutable struct FUSEparameters__ActorFluxMatcher{T<:Real} <: Paramete
     evolve_plasma_sources::Entry{Bool} = Entry{Bool}("-", "Update the plasma sources at each iteration"; default=true)
     find_widths::Entry{Bool} = Entry{Bool}("-", "Runs turbulent transport actor TJLF finding widths after first iteration"; default=true)
     max_iterations::Entry{Int} = Entry{Int}("-", "Maximum optimizer iterations"; default=500)
-    algorithm::Switch{Symbol} = Switch{Symbol}([:anderson, :newton, :trust_region, :simple, :none], "-", "Optimizing algorithm used for the flux matching"; default=:anderson)
+    algorithm::Switch{Symbol} = Switch{Symbol}([:broyden, :anderson, :newton, :trust_region, :simple, :none], "-", "Optimizing algorithm used for the flux matching"; default=:broyden)
     step_size::Entry{T} = Entry{T}(
         "-",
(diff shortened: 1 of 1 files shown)


bclyons12:
Pro: they give the same answer now!
Con: :KDEMO fails in the warmup

bclyons12:
Okay, tests are passing, mostly because of [d13dccb](https://github.com/ProjectTorreyPines/FUSE.jl/pull/890/commits/d13dccbbf17b55b1cf0ad68fd66147483e24ff40) I believe, which prevents it from erroring if something happens inside of the objective function. 

There are still some things to look into though:
1. D3D is very slow, probably because it's doing a lot of (expensive) iterations. We should reduce the default iterations significantly for a Jacobian-based method
2. DTT is giving a _ton_ of warnings.
3. Things like KDEMO are still not converging.
4. For cases that do converge, there may be even better methods than the one used.

orso82:
all checks pass! thank you for all the effort that you've put into this @bclyons12 ! 🥇

PR #918, opened 2025-06-12
Deployment scripts for NERSC Perlmutter

bclyons12:
This does everything but build the module file. I'll leave that to @ModestMC and @AreWeDreaming to take care of.

Currently running to create the full sysimage. I'll update here once it completes.


bclyons12:
@AreWeDreaming Note that, as on Omega, this builds the sysimage from the tutorial and the regression tests. So for [v0.8.5](https://github.com/ProjectTorreyPines/FUSE.jl/releases/tag/v0.8.5), this will not build anything for the time-dependent workflow. @orso82 has developed a time-dependent regression test, so the next released version would include this in the sysimage.

bclyons12:
Some things to note:
1. I had to do the install on a worker node. I’m not sure why it failed on the login node for sure. Maybe because of some memory limit I hit when building the sysimage?
2. I've made Jupyter kernels, but they need to be copied into the user's `${HOME}/.local/share/jupyter/kernels/` folder in order to be seen on https://jupyter.nersc.gov/. Even so, the kernel isn't working without the `fuse` module loaded, so I'm working on this.

PR #921, opened 2025-06-17
SOL box model 0D

orso82:


jguterl:
I suggest to add flux expension. Might be significant for large devices...

chrismarsden7:
> I suggest to add flux expension. Might be significant for large devices...

Flux expansion is included in the model

orso82:
@chrismarsden7 the ActorSOLBox is now in master.

There are a few things that I still think they can be improved. I'll open a PR for us to iterate on.

I'll close this PR and we'll come back to the integration in yet another PR.

PR #927, opened 2025-06-25
Update Perlmutter deployment

bclyons12:
- Set JULIA_DEPOT_PATH properly so user gets their own history, place to install packages, etc.
- Automate running of `julia` command on worker node, installation in scratch, then copying to `/global/common/software/m3739/perlmutter/fuse`


bclyons12:
Everything should work properly now except for Jupyter notebooks. I tried some stuff but it's not working yet. I can explain to @ModestMC and @AreWeDreaming what I tried and let them takeover

PR #932, opened 2025-06-27
Move initialization functions to QED

bclyons12:
There was too much work being done inside of ActorQED, things that other packages might want. I moved these to QED.


bclyons12:
@orso82 Please take note that I merged this. Future development on the QED initialization functions should be done inside of QED.

PR #943, opened 2025-07-29
Fix user-defined ped_to_sep ratio in ActorPedestal and add fast ion physics to ActorSimpleNB

fmunguia-wulf:
### **Change 1: Fix ActorPedestal User-Defined Density Ratios** 
ActorPedestal was defaulting sep-to-ped density ratios ($n_{e,i}^{ped}/4$) instead of preserving user-defined ratios from initial conditions. Now calculates and maintains the user's existing `sep_to_ped_ratio` from the input profile for both electrons and ions when EPED & WPED models are called.

### **Change 2: Add Fast Ion Physics to ActorSimpleNB** 
Attempted to add `IMAS.fast_particles_profiles!(dd)` call to populate fast ion density/pressure profiles. Currently creates species mismatch error with FluxMatcher actor when ActorStationaryPlasma is run. 

### **Change 3: Fix Worker Buildup in Multi-Objective Optimizer**
(I am not sure if this fix is needed, but I was interested in seeing if I could deploy a fix)

Large buildup of Julia workers was occurring when MOOP crashed, leaving orphaned processes that consumed memory but no CPU time. Added `cleanup_workers()` function to remove distributed workers on startup and error conditions. Enhanced error handling with try-catch blocks to ensure workers are cleaned up even when optimization fails. 

Example of buildup after crash and rerun: 
Screenshot 2025-07-21 at 1 46 08 PM

### DIFF
--- a/src/actors/hcd/nbi/nb_simple_actor.jl
+++ b/src/actors/hcd/nbi/nb_simple_actor.jl
@@ -263,6 +263,10 @@ function _step(actor::ActorSimpleNB)
         ion.power_inside = source1d.total_ion_power_inside
         ion.fast_particles_energy = beam_energy
     end
+    # Add this line to populate fast ion physics
+    IMAS.fast_particles_profiles!(dd)
+
+
     IMAS.unfreeze!(cp1d, :zeff)
 
     return actor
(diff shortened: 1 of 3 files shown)


src/actors/hcd/nbi/nb_simple_actor.jl
```diff
@@ -263,6 +263,10 @@ function _step(actor::ActorSimpleNB)
         ion.power_inside = source1d.total_ion_power_inside
         ion.fast_particles_energy = beam_energy
     end
+    # Add this line to populate fast ion physics
+    IMAS.fast_particles_profiles!(dd)
```
orso82, line 267: `fast_particles_profiles!()` should not be called by the `ActorSimpleNB `, but at the end of the `step()` of the `ActorHCD`. This will add fast ions independently of the NBI, IC actor being run.

Note the comment at the end of `fast_particles_profiles!()` function: **# must decide how to handle quasineutrality`. I suspect this is why this function was not hooked up.

Let's make sure to come up with a plan/system to handle quasi neutrality before merging this PR!

Issue #948, opened 2025-08-06
Fix ARC and CAT cases with updated strain limits


bclyons12:
This PR updated the ITER Nb3Sn thermal and JxB strain limits: https://github.com/ProjectTorreyPines/IMAS.jl/pull/271

This is causing the ARC and CAT tests to fail with the below error. Presumably something needs to get changed in those cases to make them compliant with the new limits.
```
== ARC ==
...
actors:  StationaryPlasma
actors:   --------------- 1/2
...
actors:   --------------- 2/2 @ 205.53%
┌ Warning: Max number of iterations (2) has been reached with convergence error of (1)[0.17, 0.103](2) compared to threshold of 0.05
└ @ FUSE ~/work/FUSE.jl/FUSE.jl/src/actors/compound/stationary_plasma_actor.jl:207
actors:  HFSsizing
actors:   FluxSwing
actors:   Stresses
┌ Warning: TF exceeds critical current: 114.03348170827343%
└ @ FUSE ~/work/FUSE.jl/FUSE.jl/src/actors/build/hfs_actor.jl:372
┌ Warning: OH exceeds critical current: 114.03346388826114%
└ @ FUSE ~/work/FUSE.jl/FUSE.jl/src/actors/build/hfs_actor.jl:372
┌ Warning: OH stresses are too high: 114.0334454332834%
└ @ FUSE ~/work/FUSE.jl/FUSE.jl/src/actors/build/hfs_actor.jl:372
┌ Warning: TF stresses are too high: 115.05721542227494%
└ @ FUSE ~/work/FUSE.jl/FUSE.jl/src/actors/build/hfs_actor.jl:372
Optimization Result
===================
  Iteration:       100
  Minimum:         4.45585
  Minimizer:       [0.372154, 0.243117, 0.403958, 0.50847]
  Function calls:  5000
  Feasibles:       0 / 50 in final population
  Total time:      3.0829 s
  Stop reason:     Maximum number of iterations exceeded.
Bool(dd.solid_mechanics.center_stack.bucked) = true
Bool(dd.solid_mechanics.center_stack.noslip) = false
Bool(dd.solid_mechanics.center_stack.plug) = true

[PL.thickness] = [0.7954958724610721]
[OH.thickness, dd.build.oh.technology.fraction_steel] = [0.25551912862452264, 0.4039575456565842]
[...]

PR #949, opened 2025-08-06
ignore engineering errors in HFSsizing in the CAT&ARC test

TimSlendebroek:


orso82:
these tests were not failing before! do we understand why they are failing now? what changed? something in VacuumFields @bclyons12 ?

bclyons12:
Based on the discussion in #948, I'm going to merge this. 

@adrianaghiozzi may "take a look at the coil thicknesses it was predicting before vs. now and double check against the radial builds we have for each case". If we can get the cases working within the limits great, though it's possible that FUSE just computes that the machines violate these limits as designed.

PR #958, opened 2025-08-11
Update TJLF compat for parametric typing

bclyons12:
Parametric typing on InputTGLF, which is used in this repo, was introduced in TJLF v1.0.16, so we require an explicit compat to that version. This will also test https://github.com/ProjectTorreyPines/TurbulentTransport.jl/pull/10


bclyons12:
@chrismarsden7 @fmunguia-wulf Once this is merged, it should solve the precompilation issue you're having.

PR #975, opened 2025-09-02
Remove unused YAML dependency

bclyons12:


orso82:
Great catch! Thank you @bclyons12

PR #983, opened 2025-09-10
Return pid list in parallel_environment for cleanup

AreWeDreaming:
This is a very simple addition that makes `parallel_environment` return the pids of the worker processes it creates. I am not 100% certain that this is the problem, but in my tests of FUSE in IRI running the code in https://github.com/ProjectTorreyPines/FuseExamples/pull/9 creates a Zombie in the end which I attribute to the  `Distributed` worker that is created but never terminated. In order to properly cleanup you need the pids from `parallel_environment`'s `Distributed.addprocs` call. This propagates them through.
Example of how it is used in https://github.com/ProjectTorreyPines/FuseExamples/pull/9

### DIFF
--- a/src/utils_begin.jl
+++ b/src/utils_begin.jl
@@ -163,6 +163,7 @@ Start multiprocessing environment
 """
 function parallel_environment(cluster::String="localhost", nworkers::Integer=-1, cpus_per_task::Int=1; memory_usage_fraction::Float64=0.5, kw...)
     if nworkers == 0
+        pid_list = []
         #pass
 
     elseif cluster == "omega"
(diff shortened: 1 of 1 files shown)


bclyons12:
@TimSlendebroek please test and review

src/utils_begin.jl
```diff
@@ -163,6 +163,7 @@ Start multiprocessing environment
 """
 function parallel_environment(cluster::String="localhost", nworkers::Integer=-1, cpus_per_task::Int=1; memory_usage_fraction::Float64=0.5, kw...)
     if nworkers == 0
+        pid_list = []
```
orso82, line 166: Not tested, but can we add type info to `pid_list`

```suggestion
        pid_list = Int[]
```

PR #989, opened 2025-09-11
Implement ActorLocking

akcayc:
** WARNING **
The actor only works for control_type=:EF (set as the default). The other control options will BREAK.

This is the first working version of the locking actor that solves an ODE system modeling a resistive plasma - resistive wall with a single rational surface (q=2)  + a resistive wall + a perfectly conducting control surface[Akcay et al PoP 2021, 2024]. The ODE system is solved as a function of 2 control parameters (currently error field and rotation rate due to instrinsic + injected NBI torque). The parallel solve--using the Distributed package-- works and has to be executed after the actor is instantiated with the inputs of the users' choice. I will leave specific instructions on how to do this part. The additional cases where the error field is fixed and another parameter, such as the tearing stability index is used as control, are NOT yet fully programmed

The actor comprises 3 parts:
   1) functions for setting up the physical and control parameters of the ODE
   2) functions for defining the RHS of the ODE and proper handling of RHS depending on the task 
   3) plotting functions

*****   RUNNING the actor ********
1) To run serially for a single ode solution (loading the default ITER or DIII-D case seems to work) and you should get a plot (still needs to be annotated) as an output:
 >> using Revise; using FUSE
>> dd=IMAS.dd()
>> ini,act = FUSE.case_parameters(:ITER; init_from=:scalars)  
>> FUSE.init!(dd, ini, act)
>> actor_out=FUSE.ActorLocking(dd,act.ActorLocking)
>> FUSE.step(actor_out)

2) To run the parallel version using Distributed (may have to change this to threaded / Orso), one follows mostly the same steps as in the serial case except a few differences with packages in the preamble, as shown below. 8 procs on my Lenovo solves the 100x100 system (1e4 ODE solves) in under a min. The output should be a scatter plot that should show solutions mostly clustered about the NW and SE corners of the Figure

>> using Revise; using Distributed
>> addprocs(num_procs)
>> using FUSE
>> @everywhere using FUSE
 then the same steps as above


bclyons12 (changes requested):
I can't get this to run, even fixing the find_q2_surface issue that I identify below.  Please provide a working example for how to run this. I'm trying:
```
dd = IMAS.dd()
ini, act = FUSE.case_parameters(:D3D, :default)
FUSE.init(dd, ini, act)
FUSE.ActorStationaryPlasma(dd, act)
act.ActorLocking.control_type = :LinStab
FUSE.ActorLocking(dd, act)
```

Some things that should be done, with people to do it
- [ ] Add tests for the automatic regressions @akcayc 
- [ ] Record results back to `dd` @akcayc 
- [ ] [Reduce dependencies and compilation time](https://docs.sciml.ai/DiffEqDocs/stable/features/low_dep) @bclyons12

PR #1012, opened 2025-10-14
Allow v1.12

bclyons12:
This reverts commit 1161c213e3aed2158a46fe940edcef14cf60e427.

This also does integration test for https://github.com/ProjectTorreyPines/TEQUILA.jl/pull/65


bclyons12:
I actually think it spends most of its time hanging after precompilation but before starting warmup_before_compile. I'm not sure what it's doing

orso82:
maybe we can switch to a single julia thread? since based on your experience Julia v1.12 can have some precompilation issues when multithreaded... https://github.com/JuliaLang/julia/issues/59861

PR #1016, opened 2025-10-23
Update Omega documentation and expose on fuse.help

bclyons12:
@orso82 Any objection?


orso82:
it was based on your feedback and @jmcclena that I switched to a more flat navigation menu, which I agree is less confusing to understand.

Perhaps you may want to try changing the "Install on specific platforms" section header

image

image

so that the OMEGA and SAGA links can appear directly in the navigation menu

PR #1064, opened 2026-03-09
Add MXH to FUSE

jmcclena:
This pull request adds higher-order MXH fits to FUSE. The MXH branch of IMAS and TurbulentTransport are required to use. 

The previous logic for Miller geometry is maintained as the default to avoid any disruptions in workflows.


tomneiser (approved):
Looks good to me!

PR #1075, opened 2026-04-06
Clamp Ti fit edge undershoot to avoid avgZ DomainError; defensive clamps for Te/ne

nanshi1177:
ActorFitProfiles crashes with a DomainError when running D3D shot 174090 with EFIT_tree="EFIT01":                                                                                                                                                                             
                                              
  DomainError with -0.007458611600149166:                                                                                                                                                                                                                                       
  log10 was called with a negative real argument...
   [4] avgZ(func, Z, Ti)  @ IMAS profiles.jl:1569                                                                                                                                                                                                                               
  [16] _step(actor::ActorFitProfiles)  @ FUSE fits_actor.jl:246
                                                                                                                                                                                                                                                                                
  Root cause                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                
  In fits_actor.jl, the Ti profile is built by a smoothing spline (IMAS.fit1d). Smoothing splines can overshoot/undershoot near the boundary when the data is sparse, producing slightly negative values at the edge (here -7.5 eV). That negative Ti is then assigned to  ion.temperature and later passed to IMAS.avgZ, which evaluates log10(Ti/1e3) to look up the average ionization state from the Zavg_z_t.dat table — and log10 of a negative number throws.                                                                                     
                                                                                                                                                                                                                                                                                
  Why EFIT01 but not EFIT02                                                                                                                                                                                                                                                     
   
  fit2d maps each CER channel (R,Z) to ρ_tor_norm using the equilibrium's ρ-interpolant. EFIT01 (auto-EFIT, magnetics-only) and EFIT02 (with kinetic constraints) produce different ψ(R,Z) → different ρ at the same channel locations. With EFIT01's mapping the edge CER points happen to land where the smoothing spline undershoots below zero. EFIT02 doesn't expose the issue, but the underlying fragility is the same.

### DIFF
--- a/src/actors/diagnostics/fits_actor.jl
+++ b/src/actors/diagnostics/fits_actor.jl
@@ -90,13 +90,12 @@ function _step(actor::ActorFitProfiles{D,P}) where {D<:Real,P<:Real}
 
     method = IMAS.NaturalNeighbours.Triangle()
 
-    # fit Te
-    itp_te = IMAS.fit2d(Val(:t_e), dd; transform=abs)
+    # fit Te in log space to guarantee positivity
+    itp_te = IMAS.fit2d(Val(:t_e), dd; transform=x -> log(abs(x)))
     for (kt, time0) in enumerate(time_basis)
         cp1d = dd.core_profiles.profiles_1d[kt]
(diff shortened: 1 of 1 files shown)


jmcclena (commented):
Is there a reason we're not fitting the log or sqrt of these profiles?

PR #1085, opened 2026-04-22
Add ZMQ coupling actor for FUSE-GSLite co-simulation

nanshi1177:
Adds ActorZMQ to enable real-time data exchange between FUSE and GSLite/GSEvolve during dynamic plasma simulations via ZeroMQ (REQ/REP pattern). 

- ActorZMQ sits inside ActorDynamicPlasma's time loop at two points:                                                                                                                                                  
   
  for each time step:      
```                                                                                                                                                                                           
      Phase 1 (current evolution):
          time_advance
          receive!(actor_zmq)    <-- get data from GSLite                                                                                                                                                             
          sources → sawteeth → j_ohmic → equilibrium → pf_active
                                                                                                                                                                                                                      
      Phase 2 (transport evolution):
          time_advance                                                                                                                                                                                                
          sources → sawteeth → pedestal → transport → equilibrium → pf_active
          send!(actor_zmq)       <-- send results to GSLite                                                                                                                                                           
```   
   This placement ensures:                                                                                                                                                                                             
       -  FUSE receives GSLite's latest flux map and actuator signals before running physics
       -  FUSE sends results after equilibrium is fully updated 
  
- Date Flow:
<img width="844" height="334" alt="image" src="https://github.com/user-attachments/assets/0c597042-e287-4176-9b7f-22f5e849495d" />
<img width="990" height="257" alt="image" src="https://github.com/user-attachments/assets/12f0aaba-615e-461d-9e31-f2a6e4580c66" />

- psizr → Equilibrium Logic                                                                                                                                                                                    
        
  The key challenge: GSLite sends a flat 2D flux array. FUSE needs either full 1D equilibrium profiles or just the plasma boundary, depending on the time step.                                                       
                  
  First time step (Method 2 — full flux_surfaces):                                                                                                                                                                    
  1. Reshape flat psizr into 2D ψ(R,Z) grid
  2. Interpolate onto FUSE's R/Z grid                                                                                                                                                                                 
  3. Find magnetic axis (find_magnetic_axis) and ψ boundary (find_psi_boundary)
  4. Run flux_surfaces(eqt, wall) which computes all 1D profiles: gm1, gm9, q, volume, area, f, j_tor, dvolume_drho_tor, etc.                                                                                         
  5. These profiles are required by QED for its first current diffusion initialization                                       
                                                                                                                                                                                                                      
  Subsequent time steps (Method 1 — boundary only):                                                                                                                                                                   
  1. Same reshape and interpolation                                                                                                                                                                                   
  2. Find magnetic axis and ψ boundary                                                                                                                                                                                
  3. Run trace_simple_surfaces to extract the LCFS boundary (R,Z points) only
  4. Write boundary to eqt.boundary.outline                                                                                                                                                                           
  5. FRESCO then uses this boundary as its target shape and recomputes internal profiles                                                                                                                              
                                                                                                                                                                                                                      
  This is efficient: Method 2 is slow but only needed once; Method 1 is fast and sufficient because FRESCO handles internal profiles after the first step.  
Design Decisions                                                                                                                                                                                                    
                                                                                                                                                                                                                      
 - I_coil stored in dd._aux only, not dd.pf_active:                                                                                                                                                                    
   We use FRESCO in shape control mode, which takes the boundary shape as input and solves for coil currents. FRESCO doesn't use coil current values from dd.pf_active in this mode. So GSLite's  I_coil is stored in dd._aux for the NN density predictor, since writing them to dd.pf_active would be pointless (FRESCO would overwrite them anyway).  

 - dd._aux (getfield(dd, :_aux)): A hidden Dict{Symbol,Any} on the dd object for runtime metadata that doesn't belong in the IMAS data structure. It's not accessible via dd._aux (dot notation),  you must use  getfield(dd, :_aux) to read/write it. Values stored here are not saved to JSON and don't persist across sessions. We use it to store GSLite signals (I_coil, Ip_avg, pr15v, gas_cal) that the NN density predictor will read later.              
                  
-  First time step output is unreliable:                                                                                                                                                                               
  On the first step, flux_surfaces runs with zero pressure and no seed current profiles, so betap, li, and p_res will be wrong. GSLite should ignore the first DataFromFUSE message.
                                                                                                                                                                                                                      
 - ZMQ connection lifecycle:
    - connect! in ActorZMQ constructor (when enabled=true)                                                                                                                                                              
    - disconnect! in finally block of ActorDynamicPlasma._step — ensures cleanup even if simulation crashes


nanshi1177:
@jmcclena Could you please check the psizr → Equilibrium part carefully to make sure we do the conversion correctly, and the logic is properly?

src/actors/zmq_actor.jl
```diff
@@ -0,0 +1,462 @@
...
+#
+# Requires: ZMQ.jl (add to Project.toml: `import Pkg; Pkg.add("ZMQ")`)
+
+import JSON
+import Interpolations
+import ZMQ
+
+@actor_parameters_struct ActorZMQ{T} begin
+    endpoint::Entry{String} = Entry{String}("-", "ZMQ endpoint (e.g., tcp://localhost:5555)"; default="tcp://localhost:5555")
+    timeout_ms::Entry{Int} = Entry{Int}("ms", "ZMQ receive timeout"; default=30000)
```
prechelg: If GSLite dies, it seems like ZMQ.recv will block forever since timeout is declared here but never applied to the socket.

Should have some sort of ZMQ.set_rcvtimeo(sock, par.timeout_ms) (and sndtimeo) in connect!

prechelg: Looks like the REQ/REP will get stuck in the wrong state if either side throws an error during the exchange.

Might want some reset logic  in there, otherwise the disconnect only happens in the outermost finally.

src/actors/zmq_actor.jl
```diff
@@ -0,0 +1,462 @@
...
+    betap = eqt.global_quantities.beta_pol
+    li = eqt.global_quantities.li_1
+    time_now = dd.global_time
+
+    # Time derivatives (0.0 on first step when prev values are NaN)
+    dt = time_now - actor.prev_time
+    if isnan(actor.prev_time) || dt <= 0.0
+        betap_dot = 0.0
+        li_dot = 0.0
+        p_res = 1e-9
```
prechelg: Why 1e-9?  Maybe a comment here to explain.

src/actors/zmq_actor.jl
```diff
@@ -0,0 +1,462 @@
...
+    # REQ: send data, then receive acknowledgment
+    ZMQ.send(actor.socket, JSON.json(msg))
+    ack = String(ZMQ.recv(actor.socket))
+    @info "ActorZMQ: sent betap=$betap, li=$li, p_res=$p_res at t=$(time_now) s"
+
+    # Store current values for next step's derivatives
+    actor.prev_time = time_now
+    actor.prev_betap = betap
+    actor.prev_li = li
+    actor.prev_psipla = _compute_psipla(eqt)
```
prechelg: It looks like you run this at line 375 (so, twice per send!).  Does caching this make sense?

src/actors/zmq_actor.jl
```diff
@@ -0,0 +1,462 @@
...
+        return Float64[]
+    end
+    eqt = dd.equilibrium.time_slice[]
+    cp1d = dd.core_profiles.profiles_1d[]
+    dens_co2 = Float64[]
+    for ch in intf.channel
+        try
+            result = IMAS.line_average(eqt, cp1d.electrons.density_thermal, cp1d.grid.rho_tor_norm, ch.line_of_sight)
+            push!(dens_co2, result.line_average)
+        catch
```
prechelg: This is a silent catch.  Consider logging the NaN failure here.

prechelg:
1. @AreWeDreaming had mentioned we wanted to use protobuf over ZMQ, but this looks like a JSON over ZMQ to me (although I must admit Julia is not my strongest language).  If this is JSON then I do worry about JSON floats losing value compared other C++ doubles on the GSLite side.

2. using context::Any and socket::Any eliminates any type inference.

3. Why is the first output unreliable and GSLite should disregard it?  Maybe there should be an explicit valid:bool field that gets sent so that GSLite doesn't need to remember that the first output is invalid.

4. aux[:zmq_*] key used Dict{Float64,...} which is keyed by dd.global_time.  Using floating point time as a dict key seems unreliable.  It looks like we're using time arrays for IP/NBI, can't we also use those for the aux signals?

5.  I think that connect! succeeds even if the endpoint is wrong.  Maybe we can handshake before setting is_connected = true

src/actors/zmq_actor.jl
nanshi1177: Previously I used 1e-9 as a safety floor because the first step has no meaningful p_res, and I  wanted to avoid sending 0.0 which could cause divide-by-zero in GSLite's circuit model. But GSlite will skip the first step anyway, 0.0 will be safe and keeps consistency with beta_dot and li_dot. I will change it.

src/actors/zmq_actor.jl
nanshi1177: Good catch — fixed.

src/actors/zmq_actor.jl
nanshi1177: log it. Also replay the last valid value when it fails.

src/actors/zmq_messages.proto
```diff
@@ -0,0 +1,47 @@
...
+// Share this file with the GSLite team to generate C++ code.
+//
+// Julia:  ProtoBuf.protojl("zmq_messages.proto", ".", "output_dir")
+// C++:    protoc --cpp_out=. zmq_messages.proto
+
+// Sent by FUSE to initiate each exchange
+message FUSERequest {
+    string status = 1;                    // "ready", "handshake"
+    double time = 2;                      // FUSE simulation time [s]
+}
```
prechelg, line 21: Might be worth setting a int32 for the schema_version here, and then checking it in your zmq_actor.

int32 schema_version = 3;

and then a following 
int32 schema_verion = 14;

inside the DataForFuse so that GSLite is returning its version too.

Then in zmq_actor.jl you should check the versions against each other:

```
const SCHEMA_VERSION = 1
_pb_send(actor.socket, FUSERequest("ready", dd.global_time, SCHEMA_VERSION))
msg = _pb_recv(actor.socket, DataForFUSE)
if msg.schema_version != SCHEMA_VERSION
  error("ZMQ schema mismatch: FUSE=$SCHEMA_VERSION, GSLite=$(msg.schema_version)")
end
```

src/actors/zmq_actor.jl
```diff
@@ -0,0 +1,496 @@
...
+        @info "ActorZMQ: GSLite signaled end of simulation"
+        disconnect!(actor)
+        return actor
+    end
+
+    # --- Update Ip in pulse_schedule so QED/FRESCO pick it up ---
+    if msg.Ip_latest != 0.0
+        ps_fc = dd.pulse_schedule.flux_control
+        IMAS.set_time_array(ps_fc.i_plasma, :reference, dd.global_time, msg.Ip_latest)
+    end
```
prechelg, line 215: this msg.Ip_latest != 0.0 concerns me a little.  If a single scalar field equals the types default (like 0 or 0.0 or false), then the encoder won't emit the value at all.  When this then gets decoded, an absent field isn't distinguished from a non absent field that happens to be actually 0.0

This shows up several times in this code, I've just called out the first one.  In this particular case, if the averaged IP was actually 0A, then this message would be valid, but not be stored in aux[:zmg_Ip_avg].

You might set the up these presence sensitive fields as optional in the .proto or use explicit bool checks for these fields.

src/actors/zmq_messages.proto
```diff
@@ -0,0 +1,47 @@
...
+    double betap = 3;                     // poloidal beta
+    double betap_dot = 4;                 // time derivative of beta_pol [1/s]
+    double li = 5;                        // internal inductance (li_1)
+    double li_dot = 6;                    // time derivative of li_1 [1/s]
+    double p_res = 7;                     // plasma resistance [Ohm]
+    repeated double dens_co2_sig = 8;     // CO2 line-integrated density [m^-2]
+}
+
+// Sent by GSLite as acknowledgment
+message Ack {
```
prechelg, line 77: This needs to be more fleshed out, just an ack doesn't pass enough information about what's going on.

```
message Ack {
  string status        = 1;           // keep for backwards compat
  bool   ok            = 2;           // false ⇒ GSLite rejected/errored
  string error         = 3;           // human-readable when !ok
  double expected_time = 4;           // optional: for desync detection
}
```

prechelg: Then you also need a bit in zmq_actor.jl

```
_pb_send(actor.socket, msg)
ack = _pb_recv(actor.socket, Ack)
if !ack.ok
  error("GSLite rejected DataFromFUSE at t=$time_now: $(ack.error)")
end
```

src/actors/zmq_messages.proto
```diff
@@ -0,0 +1,47 @@
+syntax = "proto3";
+
+// Protocol Buffer definitions for FUSE <-> GSLite/GSEvolve ZMQ communication.
+// Share this file with the GSLite team to generate C++ code.
+//
+// Julia:  ProtoBuf.protojl("zmq_messages.proto", ".", "output_dir")
```
prechelg: This command only works if you happen to be in src/actors when you run it.  Maybe that should be a comment?

src/actors/zmq_actor.jl
```diff
@@ -0,0 +1,496 @@
...
+            push!(aux[sym].values, msg.gas_cal[k])
+        end
+    end
+
+    # --- Update equilibrium from flat psizr array ---
+    if !isempty(msg.psizr)
+        psizr_flat = msg.psizr
+        nR = par.nR
+        nZ = par.nZ
+        @assert length(psizr_flat) == nR * nZ "ActorZMQ: psizr length $(length(psizr_flat)) != nR*nZ = $(nR*nZ)"
```
prechelg: I hope this assert is only here for debugging, since assert can be overridden easily in Julia

prechelg: Might consider throwing an error here if these don't match, instead of asserting

src/actors/zmq_actor.jl
```diff
@@ -0,0 +1,496 @@
...
+    io = IOBuffer(raw)
+    d = ProtoBuf.ProtoDecoder(io)
+    return ProtoBuf.decode(d, T)
+end
+
+@actor_parameters_struct ActorZMQ{T} begin
+    endpoint::Entry{String} = Entry{String}("-", "ZMQ endpoint (e.g., tcp://localhost:5555)"; default="tcp://localhost:5555")
+    timeout_ms::Entry{Int} = Entry{Int}("ms", "ZMQ receive timeout"; default=30000)
+    enabled::Entry{Bool} = Entry{Bool}("-", "Enable ZMQ coupling with external code"; default=false)
+    nR::Entry{Int} = Entry{Int}("-", "Number of R grid points for psizr reshape"; default=33)
```
prechelg: If GSLite doesn't send a 33x33 grid over (r_grid, z_grid) then the assert at line 265 kills the run

prechelg:
We should add handling for REQ/REP recovery and timeout.

the REQ socket enforces send → recv → send → recv → ...  If something breaks (because of some exception, or a caught error) and you try to send again you'll get "ZMQError: Operation cannot be accomplished in current state (EFSM)"

There are a number of places that could break this pattern.  If that happens, nothing reconstructs the socket so the coupling dies and the simulation stops.

On the timeout front, it looks like RCVTIMEO=30000 is set, so a stalled out peer would cause ZMQ.recv to raise a ZMQError after 30 seconds.  We should log this ("timeout waiting for GSLite") and retry on the next step?  maybe kick off a controlled shutdown on timeout?  

----------

I also noticed there isn't a ZMQ actor test?

jmcclena (approved):

PR #1088, opened 2026-04-23
Allow setting cache location through environment variable for DIII-D case

AreWeDreaming:
This is a draft because I have not tested the code yet.

I refactored the DIII-D data loading and added an environment variable that controls the location of the local cache of DIII-D data. There are two reasons for this:
- For the digital twin batches we will have a pathfinder run that runs before we send the batch. It will trigger a cache miss and then pull in the data from OMAS and OMFIT as usual. Once we run these batches we must not trigger OMAS/OMFIT and therefore we need to have a fixed location of the cache. This location will be bind mounted into the container Currently the cache folder would have been `/tmp` which is not a networked directory.
- For the container build of FUSE I need to run the `run_time_dependent_d3d.jl` script in FuseExamples to track which functions need to be dumped into the sysimage. I do not want to deal with the OMAS/OMFIT component when assembling the sys-image/container, hence I will dump the required DIII-D data into the container so that FUSE can grab it from there.

### DIFF
--- a/deploy/perlmutter/base.lua
+++ b/deploy/perlmutter/base.lua
@@ -31,7 +31,7 @@ else
 end
 
 -- Environment variables for data fetching
-setenv("FUSE_D3D_CACHE_DIR", "/pscratch/sd/d/" .. os.getenv("USER") .. "/FUSE_D3D_data_cache")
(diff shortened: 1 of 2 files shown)


AreWeDreaming:
Tested the old pathway where there is no cache data successfully and the d3d data files showed up in the folder from the environment variable. I will now proceed to test the pathway where we get a cache hit by attempting the container build.

PR #1091, opened 2026-05-11
Akcay branch for MiRACL IMAS-> PMI workflow, using CHEASE+MARS

akcayc:
This branch drives CHEASE + MARS to build a workflow for MiRACL. Both CHEASE and MARS enter as external codes. 

CHEASE runs when par.run_equilibrium = true (default), writes an EXPEQ data file from dd, then overwrites the default namelist entries with user choices, then executes. The actor outputs the standard CHEASE outputs, including OUTVMARS and OUTRMARS required by MARS, and the updated namelist _datain_ among. Rerun capability is provided with parameter restart_equilibrium=true (default is false). This equilibrium has been benchmarked against EFIT02 163303.3000 (things agree to within CHEASE's modification of the original dd)

MARS is run when par.run_MHD=true (default). It's currently only been tested as an eigenvalue solver for an external kink and RWM, just to test the mechanics.

To **TEST**, run runtests_actors. Several test cases are considered. As these are external codes, the user MUST specify the absolute PATHS to the executables for both codes (the default paths are set to CHEASE and MARS paths under /fusion/projects/codes/mars on Omega), otherwise things will BREAK. But the testing should produce messages consistent with this.

There is currently NOTHING being saved back to dd, which will come next.

TO-DO for future revision (NOT for this PR)

- Test basic REORBIT drift orbit tracing capability for a single particle
- Scale up the orbit tracing to many particles (migrate to NERSC?)
- Write back to dd, especially those quantities relevant for MiRACL
- Consider FUSEfying REORBIT to // the orbit calculations over a GPU
- Benchmark rMHD as well as rmHD+ drift kinetic cases on published MARS results.


torrinba:
@akcayc is your TODO list planned for this PR or future ones? If it's the former you should probably convert this to a draft/WIP

akcayc:
That TO-DO list is for the future revisions, I should make that clear.

src/actors/equilibrium/chease_actor.jl
```diff
@@ -86,9 +86,11 @@ function _step(actor::ActorCHEASE)
     # scalars
     Ip = eqt.global_quantities.ip
     Bt_center = eqt.global_quantities.vacuum_toroidal_field.b0
+    Bt_axis = eqt.global_quantities.magnetic_axis.b_field_tor
```
jmcclena: Looks like the edits to chease_actor.jl aren't used

src/actors/stability/mars_actor.jl
```diff
@@ -0,0 +1,900 @@
...
+    _time::Float64 = NaN
+    do_plot::Entry{Bool} = act_common_parameters(; do_plot=false)
+    eq_type::Switch{Symbol} = Switch{Symbol}([:CHEASE, :TEQUILA], "-", "Type of equilibrium to use: :CHEASE or :TEQUILA"; default=:CHEASE)
+    EQDSK::Entry{Bool} = Entry{Bool}("-", "Enable EQDSK"; default=false)
+    MHD_code::Switch{Symbol} = Switch{Symbol}([:MARS_Q, :MARS_F, :MARS_K], "-", "MHD code to use: :MARS or :MARS_F"; default=:MARS_F)
+    chease_exec::Entry{String} =
+        Entry{String}("-", "Path to CHEASE executable"; default="/fusion/projects/codes/mars/CHEASE/chease.x")
+    offset::Entry{Float64} = Entry{Float64}("-", "Offset for first wall (RW) in meters"; default=0.2)
+    n_points::Entry{Int} = Entry{Int}("-", "Number of points for plasma boundary and surrounding walls "; default=301)
+    tracer_type::Switch{Symbol} = Switch{Symbol}([:ORBIT, :REORBIT], "-", "Type of tracer to use: :ideal or :realistic"; default=:REORBIT)
```
jmcclena, line 220: I'd think there'd be utility in running without the tracer. Add a `:none` option and make it default?

src/actors/stability/mars_actor.jl
```diff
@@ -0,0 +1,900 @@
...
+end
+
+
+"""
+  Writes a EXPEQ file for CHEASE given dd and mode number
+    This function will create the necessary input file for the CHEASE equilibrium solver.
+        NWBPS: Number of wall boundary points
+        NSTTP: Number of steps in pressure and current profiles
+"""
+function write_EXPEQ_file(dd::IMAS.dd, par; wall_data=nothing)
```
jmcclena: These CHEASE routines evenutally be moved to CHEASE.jl (probably before merging)

src/actors/stability/mars_actor.jl
```diff
@@ -0,0 +1,900 @@
...
+    # Offset curve
+    X = xv .+ d .* Nx
+    Y = yv .+ d .* Ny
+
+
+    return X, Y
+end
+
+
+function julia_grep(
```
jmcclena, line 1198: Is this just used for debugging?

test/runtests_actors.jl
```diff
@@ -15,3 +15,74 @@ using Test
     act.ActorTGLF.model = :TGLFNN
     FUSE.ActorFluxMatcher(dd, act)
 end
+
+@testset "Mars Actor CHEASE run" begin
```
jmcclena, line 19: We probably need to add logic to make sure this doesn't crash the automatic GitHub testing.

src/actors/equilibrium/chease_actor.jl
akcayc: I fixed the bugs I accidentally introduced in March. Sorry for confusing you (and myself) with my earlier response

src/actors/stability/mars_actor.jl
akcayc: The tracer option is really not being used. By default, the workflow is run without the tracer, and the parameter _num_orbits_ is supposed to turn ON REORBIT, assuming the mars input file in workdir contains REORBIT namelist options. I originally wanted a general workflow in which I could swap equilibrium, MHD stability codes, particle tracers etc, by calling up different actors. Since then, I locked myself into CHEASE + MARS +REBORBIT.

src/actors/stability/mars_actor.jl
akcayc: This is a step that the current MARS-CHEASE workflow does to simply display q_0, q_min, q_edge after CHEASE runs as a quick check/debugging tool. So, I copied it and rewrote it this way. It is not elegant. How can I improve it?

I need the CHEASE output to update the dd, at least for plotting purposes, as chease_actor does. I will probably work toward modifying CHEASE package to accommodate my needs in the coming days/weeks to avoid code duplication, like @torrinba said

src/actors/stability/mars_actor.jl
akcayc: Yes, I think that is the best course of action moving forward, to leverage what is already there. But I will have to modify the source in the CHEASE package to accommodate the needs of Martian CHEASE.

test/runtests_actors.jl
akcayc: I don't have an elegant solution to automagically get the paths to the exec. populated correctly. Again, I am looking at the contents of the CHEASE package (again) to see how it is done and how that can be leveraged (after appropriate mods)

src/actors/stability/mars_actor.jl
akcayc: All  CHEASE routines in the Martian actor were moved to the CHEASE package.

PR #1106, opened 2026-06-12
(compat): Widen JSON compat to v1

mgyoo86:
## Summary 
Widen `JSON` compat to "0.21, 1" to allow using JSON version 1.x.


mgyoo86:
@jmcclena 
This PR now passes the CI, so I think it is safe to merge.

PR #1112, opened 2026-06-19
ActorLocking: reduced-order model of locking to a resonant error-field and locking probability estimate with NN probability classifier

akcayc:
### **Physical model**

`ActorLocking` solves a set of coupled nonlinear ODEs describing the resistive-MHD interaction between a tearing mode (TM) and a resistive wall (RW) or ideal wall (IW) in the presence of plasma rotation and external error fields applied at the IW. The model captures the bifurcation between a rotating (unlocked) and a locked state as a function of two control parameters: the applied rotation frequency Ω₀ and a swept second control. Inductances and stability indices are derived directly from the FUSE `dd` equilibrium. ODEs are solved on a 2D grid of a pair of control parameters, which are proxies for experimental control or dynamically changing plasma parameters, for example Ω₀ for DIII-D can represent the NBI torque near the q=m_pol/n_tor rational surface

### **Control scenarios**

Three control types are supported via `par.control_type`:

- **`:EF`** — sweep the error field amplitude (x-axis in physical units: Gauss) at fixed stability index; captures the classical error-field-induced locking threshold
- **`:LinStab`** — sweep the linear stability index Δ′ (must be negative for a stable tearing mode) to map how proximity to marginal stability shifts the locking boundary
- **`:NLsaturation`** — sweep the nonlinear saturation parameter α; x-axis displayed as 1/α (stronger saturation → smaller 1/α)

### Tasks

The actor is driven by `par.task`:

| Task | Description |
|---|---|
| `:single_case` | Integrate one ODE trajectory at a user-specified (C1, C2) point |
| `:solve_system` | Parallel grid scan over (Ω₀, control) space; results saved to disk |
| `:calc_prob` | Train a neural-network binary classifier (locked/unlocked) on saved ODE results; model saved to disk |
| `:evaluate_probability` | Load a previously trained NN and evaluate locking probability |
| `:transfer_learning` | Run a focused grid scan on a new equilibrium, then fine-tune the last layer of the base NN model |

### ModeLocking package

All physics, ODE integration, NN training, persistence, and plotting are encapsulated in the standalone [`ModeLocking.jl`](https://github.com/ProjectTorreyPines/ModeLocking.jl) package (registered in FuseRegistry v0.2.0). `ActorLocking` is a thin FUSE wrapper that populates `ODEparams` from `dd`, drives the task workflow, and provides plot wrappers that pass physical normalization scales (b₀, t₀, m_pol) for axis unit conversion.

### Testing

The actor test suite lives in `test/runtests_locking.jl` and covers all tasks:

- **`single_case`** — verifies a single ODE integration runs without producing a results grid
- **`solve_system_EF`** — grid scan with `:EF` control; checks ODE solution dimensions, k-means classification labels, analytic bifurcation bounds shape and finiteness
- **`solve_system_LinStab`** — grid scan with `:LinStab` control over a negative Δ′ range `[-3, -1]`
- **`solve_system_NLsaturation`** — grid scan with `:NLsaturation` control; verifies bifurcation bounds are suppressed when NL saturation is active
- **`calc_prob_no_ode_rerun`** — verifies `:calc_prob` loads saved ODE results from disk and trains the NN without re-solving the ODEs
- **`transfer_learning`** — seeds a base NN model, runs a focused grid scan on a restricted control range, fine-tunes the last layer, and checks the output probability is in `[0, 1]`

The `ModeLocking.jl` package has its own independent test suite (`ModeLocking/test/runtests.jl`) covering types, ODE solvers, normalization, grid scans, the NN classifier, persistence round-trips, and all plotting functions.

### What's been completed since the last PR
- finalize() writes back to dd
- task=:eval_prob does an evaluation of the locking probability for M-many points in the control space, and places these points in the prob. plot
- Error field is specified in Gauss, then converted to whatever dimensionless form the ODEs use
- control_type=:LinStab is specified in terms of the full RP-RW stability index (NO longer just the TM index), so the upper limit in the range is now 0 - epsilon. The eval_prob operational points for this control type is specified in terms of the observed n=1 magnetic perturbation amplitudes at the q=2 rational surface, again in Gauss. The actor maps these amplitudes back to the stability index for the evaluation
- There is a demo_locking.jl I also wrote that provides a nice demo in addition to the standard testing.
- All the tests passed 

 

### Future work

0. Might need a calibration of model's psit, psiw, eps against int, ext B-perp and error field
1.  Finish Transfer learning implementation
2.  Coupling to the MARS actor to provide an EF at some r_c based on I-coils
3. Generalization to multiple resonant layers (e.g. coupled (2,1)+(3,1) tearing modes)
4. Island width calculation and self-consistent estimation of the nonlinear saturation parameter α from observed island sizes
5. Coupling the modified Rutherford equation (MRE) to the system to compute the Maxwell torque over the full island width
6. Continuum resonance torque or neoclassical toroidal viscosity (NTV) approximations for the momentum source term


akcayc:
@adrianaghiozzi it is finally READY to be reivewed.

adrianaghiozzi:
@akcayc and I just got off a call about this PR, will summarize the discussion here for posterity: 

- For a naive user to get the most out of this model, more things need to be written out to dd 
  - Specifically a simplified locking probability or tolerable error field for a given machine design 
  - The left boundary between the locked region and the hysteretic region could be written as an array to dd 
  - This will enable creation of a plot(dd.mhd_linear) and a do_plot parameter so that a user can see a simplified version of the plot without having to save them all to file 
- We want to maintain self-consistency with dd 
  - [CHECK] Don’t force a source torque for a machine that doesn’t have externally driven NBI
  - Or if introducing a source torque that doesn’t come from dd.core_sources, set no default for par.source_torque so that the user is forced to knowingly choose a value 
  - If par.r0 is usually the minor radius then the actual minor radius can just be pulled from dd.equilibrium.time_slice[].boundary.minor_radius
- [CHECK] Remove the locking_actor_changelog.txt file
- [CHECK] Remove :Monte_Carlo task given that it is not implemented yet

akcayc:
@adrianaghiozzi , ready once again for your review. I updated the word doc on google

adrianaghiozzi:
From my perspective this is good to go 🥳 

The only remaining thing that I think would really tie it all together is if we had some way to store a minimal representation of all that this outputs in dd.limits. @akcayc and I have spent a fair amount of time talking about this but haven't really decided on the best solution. 

In a perfect world, one would have ActorPlasmaLimits call ActorLocking for the full :solve_system task, and write the locked_fraction that results from ActorLocking to dd.limits.model[i].fraction. I think that would be really cool as a sort of 0D picture of how susceptible a given case is to locking. We didn't end up adding that, though, because the full solve_system takes ~1 minute, while the rest of the ActorPlasmaLimits default models take < 1 second and are really useful for the optimization loop (so a minute per call is way too long). Another option, maybe the most feasible, would be to just have ActorLocking write locked_fraction to dd.mhd_linear, and then have ActorPlasmaLimits do something like "if dd.mhd_linear...locked_fraction != nothing, dd.limits.model.fraction = dd.mhd_linear...locked_fraction" - basically just copy it over. Cihan also went to the trouble of adding the task = :bounds, which can rapidly (~ms) give you the boundary between the locked and hysteretic regions, so another option would be to extend dd.limits to accept some kind of stability_boundary vector and write the vector of values that define the boundary there. 

All that to say - this actor is totally functional as-is and imo should be merged (Columbia folks are interested in making use of it already!). My thoughts here are mostly for posterity should anyone decide to revisit writing out these results to dd.limits in the future. 

@jmcclena Would love your thoughts on this if you have any! And otherwise this can be merged (I don't have merging powers so that's all you)

adrianaghiozzi (approved):

akcayc:
Unfortunately, locking fraction itself may be a misleading number because it will change when you change the dimension of the 2D control space (if you expand it in the direction of the error field, for example, you will include more of the strictly locked states, thereby increasing the overall locked fraction. After the probability is calculated, one can simply evaluate it (task=:prob_eval) at chosen places in the control space by specifinyg op_C1, op_C2 vectors in par. That would be very quick, and produce either a scalar or a small vector

akcayc:
For the limits actor, one can do a scalar evaluation on the :bounds result from the locking actor for a chosen rotation frequency at the mode rational surface. That would yield the penetration threshold EF

PR #1128, opened 2026-07-16
Added a version of QLGYRO that submits jobs via slurm using the Turbu…

Seanlyons316:
Added a version of QLGYRO that submits jobs via slurm using the TurbulentTransport.run_qlgyro function a
nd integrated that into flux_calculator_actor.jl.


jmcclena:
@Seanlyons316 Have you tested this? Also, I don't think there is a need to have a separate script for a Slurm run.

PR #1148, opened 2026-08-05
Support sysimage-less arm64 container builds (FUSE_PRECOMPILE_WORKLOAD=none)

tomneiser:
The published `ghcr.io/projecttorreypines/fuse` image was amd64-only, so Apple Silicon users ran it under emulation. Building the arm64 image surfaced a hard platform limit: **no aarch64 FUSE sysimage can link**. The serialized sysimage data blob (~3.2 GB, measured as the `.ldata` section of the published amd64 `sys_fuse.so`) overflows the ±2 GiB span of the `R_AARCH64_PREL32` relocations Julia emits:

```
/tmp/jl_...-o.a(text#0.o):(.rodata+0x36dd0): relocation truncated to fit:
R_AARCH64_PREL32 against symbol jl_globalYY... defined in .bss
collect2: error: ld returned 1 exit status
```

All variants fail identically (verified on an M2 Max, FUSE v1.1.6 / Julia 1.11.7, with up to 185 GB of virtual memory — this is not a resource issue):
- `generic;apple-m1,clone_all` (the current workflow setting) — fails
- single `generic` target — fails (`.text` size is not the binding constraint)
- `FUSE_PRECOMPILE_WORKLOAD=light` — fails

x86_64 escapes only because the medium code model places the blob in the `.ldata` large-data section beyond `.bss`; aarch64 has no equivalent, so the blob sits inside the `.rodata`→`.bss` span. Upstream context: https://github.com/JuliaLang/PackageCompiler.jl/issues/1019

- **`install_fuse_container.jl`**: new `FUSE_PRECOMPILE_WORKLOAD=none` mode — full package install + pkgimages precompile, no `create_sysimage`
- **`Containerfile`**: the `fuse` launcher falls back to plain `julia` when `sys_fuse.so` is absent, and unsets `JULIA_CPU_TARGET` there so the runtime JIT targets the native CPU rather than the generic build-time target
- **`container.yml`**: arm64 matrix leg switches to `cpu_target: generic` + `workload: none`. Bonus: without the ~98 GiB sysimage emission, the arm64 leg fits standard `ubuntu-24.04-arm` runners

Trade-off for arm64 users: `import FUSE` takes tens of seconds (pkgimages) instead of ~5 s (sysimage); compute performance after load is native and unaffected.

## Already published

The multi-arch `:v1.1.6` / `:latest` manifests (pushed 2026-08-04) already include an arm64 image built exactly this way, alongside the untouched omega-built amd64 image (per-arch tags `:v1.1.6-amd64` / `:v1.1.6-arm64`). Verified on Apple Silicon: `docker pull` resolves `linux/arm64` and `fuse -e 'import FUSE'` reports 1.1.6 on aarch64.

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -4,12 +4,20 @@ name: container
 # linux/amd64 and linux/arm64 (native Apple Silicon) built in parallel and
 # combined into one multi-arch tag.
 #
-# MANUAL DISPATCH ONLY — not viable on standard GitHub-hosted runners: the
-# sysimage object emission peaks at ~98.5 GiB RSS (measured on omega,
-# 2026-07-26), far beyond their 16 GB (swap and workload trimming both fail;
-# runs 30170851378, 30177176119, 30185669953). Requires 32-core/128 GB larger
-# runners to ever run in CI. Until then, release images are built on omega
-# and pushed with deploy/omega-container/publish_ghcr.sh.
+# MANUAL DISPATCH ONLY — the amd64 leg is not viable on standard GitHub-hosted
+# runners: the sysimage object emission peaks at ~98.5 GiB RSS (measured on
+# omega, 2026-07-26), far beyond their 16 GB (swap and workload trimming both
+# fail; runs 30170851378, 30177176119, 30185669953). Requires 32-core/128 GB
+# larger runners to ever run in CI. Until then, amd64 release images are built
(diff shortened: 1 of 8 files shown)


jmcclena:
@tomneiser, what are all these changes to the init that are hiding in this pull request?

tomneiser:
These are robustness fixes from running time-dep simulations on recent high betap experiments for Siye

PR #1149, opened 2026-08-06
Harden install scripts: enforce Julia >= 1.11, auto-recover from git ownership errors, add full Windows install CI

tomneiser:
## Summary

Three related improvements to the FUSE install experience:

### 1. Enforce the minimum supported Julia (1.11)

- Adds `julia = "1.11"` to the `[compat]` section of `Project.toml`, matching the oldest Julia in the regression matrix (`.github/workflows/runtests.yml`).
- `scripts/install_fuse_common.sh` and `scripts/install_fuse_windows.ps1` now verify the detected Julia version. When it is too old and `juliaup` is available, they automatically switch to the `release` channel and re-check; otherwise they fail with a clear remediation message. Compatible Julia installations and juliaup channels are left untouched.
- `docs/src/install.md` documents the requirement.

### 2. Auto-recover from libgit2/git ownership errors (`GitError(Code:EOWNER, ...)`)

libgit2's CVE-2022-24765 mitigation refuses to open git repositories cloned under a different ownership context — on Windows this typically happens after an earlier install ran from an elevated ("Run as administrator") PowerShell.

- `scripts/install_fuse_julia.jl` wraps the registry/Pkg operations in `with_registry_repair`: on an ownership error it deletes the offending `FuseRegistry`/`General` registry clones (disposable caches), re-adds them, and retries once.
- The bash and PowerShell installers handle the analogous git ownership rejection for `FuseExamples` — since that clone may hold user work, it is marked with `git config --global --add safe.directory` instead of being deleted.
- `Clone-FuseExamples` in the PowerShell script now checks `$LASTEXITCODE` after `git fetch`/`reset`/`clone` and fails with actionable messages instead of continuing silently.
- A troubleshooting entry for the EOWNER error is added to `docs/src/install.md`.

### 3. Full Windows install CI

- Adds a `windows-full-install` job to `.github/workflows/test_install_windows.yml` that runs `scripts/install_fuse_windows.ps1` end-to-end on `windows-latest` with the latest stable Julia — covering registry add, `Pkg.add("FUSE")` + precompile, Revise, fusebot, the conda env, IJulia kernel registration, and the FuseExamples clone — then verifies that `using FUSE` loads and that Jupyter kernels are registered. The physics-heavy smoke test and notebook solve are skipped (`FUSE_SKIP_SMOKE`/`FUSE_SKIP_VERIFY`) to keep the job within CI limits; those paths are covered by `runtests` on Linux/macOS.
- `scripts/install_fuse_julia.jl` gains an `abspath(PROGRAM_FILE) == @__FILE__` guard so it can be `include`d without side effects.


jmcclena:
@tomneiser was there something that came up that required adding `julia=1.11`? I thought having it in IMASdd.jl would already enforce it.

tomneiser:
Yes, for users with Julia 1.10, the resolver doesn't refuse to install FUSE because of IMASdd — it simply downgrades IMASdd to 8.4.0 (which IMAS's "8.3" bound happily admits) and keeps going

PR #1152, opened 2026-08-10
Solps chis

jmcclena:
This pull request changes SOLPSNN actor to use diffusivities and chis consistent with the dd.

In testing, the density diffusivity was well out of bounds for the ARC, which makes me a bit nervous. 
Requires IMAS.jl 7.3.0


tomneiser:
Can you try running SOLPS-NN on the ITER example case, please? Since it was definitely trained on that input space, any out-of-bounds warnings imply improper scaling of the inputs before feeding them to the NN

tomneiser (approved):

PR #1157, opened 2026-08-18
ActorTEQUILA: validate the solve and fall back to Picard when VEQ stalls

tomneiser:
Fixes the sporadic KDEMO whole_facility CI failure: the VEQ inner Newton occasionally stalls on marginal inputs (warm-started shot whose profiles moved a lot between StationaryPlasma iterations; FP differences across CI runners flip marginal cases). TEQUILA only warns and returns as long as its surfaces stay nested, but the unconverged psi map is non-monotonic, so flux-surface tracing yields degenerate surfaces (r_outboard == r_inboard) that blow up two actors later in Sauter bootstrap (AssertionError: a_eq .> 0.0).

- assert_tequila_shot_valid: fail fast right after any TEQUILA solve if the nodal psi is non-finite/non-monotonic or the surfaces non-nested.
- :veq solve failures (invalid result or thrown error) now fall back to the Picard solver from a fresh boundary shot, with a warning carrying the original error, instead of poisoning downstream actors. This also makes the :veq default usable with older TEQUILA versions that lack veq_solve! (falls back instead of UndefVarError).
- ActorEquilibrium backstop: after IMAS.flux_surfaces, error with a clear solver-attribution message on degenerate traced surfaces for any equilibrium model.

Validated: KDEMO init + validator unit checks + implicit fallback pass with dev TEQUILA (no veq_solve!); genuine veq-solved KDEMO shot passes the validator in the v1.2.0 container (new TEQUILA), so no false positive.

### DIFF
--- a/src/actors/equilibrium/equilibrium_actor.jl
+++ b/src/actors/equilibrium/equilibrium_actor.jl
@@ -151,9 +151,11 @@ function _finalize(actor::ActorEquilibrium)
         # surfaces, which then blows up far away in downstream physics (e.g.
         # Sauter bootstrap asserting on minor radius); fail fast here instead
         a_eq = eqt.profiles_1d.r_outboard .- eqt.profiles_1d.r_inboard
-        if !all(isfinite, a_eq) || any(x -> x <= 0.0, @views a_eq[2:end])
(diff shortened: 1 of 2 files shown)


jmcclena:
@tomneiser I thought that fuse refits the boundary every time it goes through ActorEquilibrium. Is veq ever run in this case?

PR #1161, opened 2026-08-22
Fixed some issues with the GSLite coupling

prechelg:
I had to make some changes to get the fuse/gslite/vpcs happy in the container together, these are robustness changes around the grid sizing.


tomneiser (approved):
Looks good to me and @nanshi1177, thanks!

nanshi1177 (commented):
Overall the changes make sense and improve robustness. The ONNX thread cap is set up correctly, and on the ZMQ side the clock fix is the most important one. ActorDynamicPlasma sets global_time on its own time grid, and the old code was overwriting it with GSLite's slightly offset sim_time.

PR #1163, opened 2026-08-25
ActorEPED: select the EPED-NN model, default to the new ensemble

TimSlendebroek:
Wires `ActorEPED` to the new `EPED1NNensemble` and makes the choice of EPED-NN a parameter, defaulting
to the new model.

Part of a three-repo change, all branches named `eped_nn_ensemble` so that CI picks up the matching
branch in each repo. **Merge order: EPEDNN first, then IMAS, then this PR** (compat here requires
`EPEDNN = "1.1"` and `IMAS = "7.4"`).

| repo | PR | what it adds |
|---|---|---|
| **EPEDNN.jl** | https://github.com/ProjectTorreyPines/EPEDNN.jl/pull/9 | `EPED1NNensemble` + `data/EPED1NNensemble.bson`, and a model-agnostic API |
| **IMAS.jl** | https://github.com/ProjectTorreyPines/IMAS.jl/pull/314 | `pedestal_poloidal_beta` and `b_field_poloidal_average` |
| **FUSE.jl** | https://github.com/ProjectTorreyPines/FUSE.jl/pull/1163 | `act.ActorEPED.nn_model`, defaulting to the new model |

## What this changes

- `act.ActorEPED.nn_model` = `:variable_nesep_ratio` (default, the new ensemble) or
  `:fixed_nesep_ratio` (the original single network, trained at a fixed `ne_sep/ne_ped ≈ 0.25`).
  The actor only fills the inputs from `dd`, clamps the height to the separatrix pressure and writes
  the result; everything model-specific lives in EPEDNN and IMAS.
- `ActorEPED.σ_frac` is the uncertainty of the pedestal height, which `ActorPedestal` writes to
  `dd.summary.local.pedestal.t_e.value_σ` and `t_i_average.value_σ` (only when the model has one).
- `run_EPED!` fills the two new inputs, so it takes `Te_sep`; `run_EPED` returns
  `(pped, wped, σ_frac)`, and `ActorEPEDprofiles` goes through the same model as `ActorEPED`.
- The width is now evaluated on the pedestal pressure that is actually used, so it stays consistent
  when the separatrix pressure sets the height.
- `pedestal_poloidal_βn` uses the new `IMAS.b_field_poloidal_average` instead of rebuilding it inline.
- `act.ActorEPED.only_powerlaw` (default `true`) applies to the single network only: the ensemble is
  always evaluated with its network correction, since the disagreement between the networks — which
  is what its uncertainty is made of — is exactly what dropping it would throw away.

## Does switching the default disturb anything?

Each case initialized with `FUSE.init`, then run through `ActorPedestal(model=:EPED)` with each model:

| case | R [m] | Ip [MA] | p_ped old [kPa] | p_ped new [kPa] | Δ | σ new | in distribution |
|---|---|---|---|---|---|---|---|
| KSTAR | 1.78 | 0.80 | 7.16 | 8.30 | +15.9% | 4.7% | yes |
| DTT | 2.19 | 5.35 | 97.09 | 95.58 | -1.6% | 4.8% | yes |
| SPARC | 1.85 | 8.70 | 368.14 | 394.64 | +7.2% | 60.1% | **no** (bt outside training bounds) |
| ITER | 6.19 | 15.00 | 63.05 | 66.09 | +4.8% | 1.2% | yes |
| FPP | 4.90 | 7.94 | 65.17 | 70.09 | +7.5% | 5.0% | **no** (σ = 5.0%) |
| KDEMO | 6.83 | 12.72 | 86.21 | 86.55 | +0.4% | 5.1% | **no** (a outside training bounds) |
| NSTX | 0.85 | 0.80 | 4.17 | 0.92 | -77.9% | 134.4% | **no** (r outside training bounds) |

Reading this table:

- **The two models agree within a few percent on the mainstream cases.** FUSE's density profiles all
  come out at `ne_sep/ne_ped ≈ 0.26`, exactly the condition the original database was generated at,
  so the two extra inputs sit near their old fixed values and what is left is the effect of the new
  ITER-scale training data.
- **SPARC and NSTX are outside the training box of *both* models** (`bt = 12.2 T` against a maximum
  of 8.0 T, `r = 0.85 m` against a minimum of 1.5 m). The difference is that the new model says so,
  with 60% and 134% uncertainty and `in_distribution = false`; the old network returns a number with
  no indication that it is extrapolating. NSTX is also a reminder that EPED is not the pedestal model
  to use for a spherical tokamak in the first place (`act.ActorPedestal.model = :analytic` is).
- **KDEMO and FPP land just over the threshold** — `a = 2.1 m` a hair outside the training bounds, and
  an ensemble spread of 5.0%, the 99th percentile of the spread over the training set. Both
  predictions are within 0.4% and 7.5% of the old ones; the flag says "at the edge", not "wrong".

One ITER case run all the way through `ActorStationaryPlasma`:

| quantity | old | new | Δ |
|---|---|---|---|
| Pfusion [MW] | 145.2 | 142.9 | -1.6% |
| Qfusion | 1.661 | 1.635 | -1.6% |
| Paux_tot [MW] | 87.4 | 87.4 | +0.0% |
| Te0 [keV] | 16.04 | 15.86 | -1.1% |
| Ti0 [keV] | 15.29 | 15.10 | -1.2% |
| ne0 [m⁻³] | 9.98e19 | 1.00e20 | +0.6% |
| βn | 1.135 | 1.123 | -1.0% |
| τe [s] | 2.092 | 2.086 | -0.3% |

Both converge and the converged profiles sit on top of each other: the +4.8% pedestal pressure of the
first iteration is largely absorbed by the transport solution. The point of this run is not the
numbers but that switching the default model does not disturb the workflow.

Full validation write-up with figures and the scripts that produced these tables:
`eped/pr_validation/PR_NOTES.md` of the iter_claude repo.


TimSlendebroek:
Don't merge this verbatim yet because I made the new EPED-NN model the default. Also merge together with changes in EPEDNN.jl and IMAS.jl at the same time otherwise it breaks. Also here is a fun place to play around with the new EPED-NN in ITER parameter space https://iter.fuseexplorer.com/EPED

PR #1164, opened 2026-08-27
Added transpose to the psi matrix that shows up from gslite (row majo…

prechelg:
…r vs column major)

I also added a test to catch this in the future.


smithsp (approved):
This all looks ok to me.

tomneiser (approved):
Looks good, thanks @prechelg!

PR #1172, opened 2026-09-04
Replace ResNet pedestal NN with fuse29 Mamba-2 ONNX predictor

tomneiser:
## Summary

Swaps the `act.ActorPedestal.ne_from = :nn_predictor` backend from the five-bundle `edensfit89` ResNet ensemble (32 z-scored channels + a 50-shot MSE history) to **fuse29**, a 29-input Mamba-2 state-space pedestal predictor for DIII-D published at [`SCS-Lab/FUSE29-Pedestal-model`](https://huggingface.co/SCS-Lab/FUSE29-Pedestal-model) (source: [NoahH72/FUSE-Pedestal-Predictor-Mamba-ONNX](https://github.com/NoahH72/FUSE-Pedestal-Predictor-Mamba-ONNX)).

- Raw physical units in (normalisation is inside the graph), physical units out, one tick per 50 ms, no shot-history / aux / plasma-response inputs (`pohm`, `ip`, `ipspr15v` are gone by design).
- Nine heads instead of one: `ne`, `te_ped`, `ti_ped`, `t_rot_ped` at ρ_tor = 0.85 (valid in every regime), four H-mode-only tanh-fit pedestal heads, and an `hmode` probability.
- Stateful: the step graph carries a ~3 MB recurrent state that must be zero at shot start and advanced exactly once per 50 ms of simulated time.

The old ResNet path is removed outright (no compatibility switch).

## What changed

**`src/actors/pedestal/nn_predictor.jl`** (rewritten)
- `Fuse29NN` loads one seed's step + sequence graphs and asserts actuator/head names and order and graph I/O names against `model_config.json` at load time.
- Thread-capped ORT session loader and stdlib-only HuggingFace auto-fetch kept; default fetch is seed 0 only (~90 MB, was ~700 MB). `FUSE_PEDESTAL_NN_SEED` selects another seed. Default artefact dir is `<FUSE>/../pedestal-predictor-mamba-onnx-fuse/artifacts` or `FUSE_PEDESTAL_NN_DIR`.
- `fuse29_step` uses the ONNX Runtime C API with the state in ORT-native (reversed-dims) layout — the generic ONNXRunTime.jl path permutes the state twice per tick and was ~4x slower.
- `Fuse29Tracker` / `fuse29_advance!`: exactly one model tick per 50 ms of simulated time (catch-up, zero-order hold, re-step from the committed state on repeated/iterated steps, reset on rewind) plus a Schmitt-trigger H-mode gate.
- `fuse29_check_units` / `fuse29_in_distribution` ported from the model's Python runtime; unsupplied channels are filled with the training **median** (the mean vector is jointly unphysical: signed bimodal Bt).

**`src/actors/pedestal/pedestal_actor.jl`**
- `fuse29_predict!` replaces the ResNet call and the 0.35 / 2-tick debounce; the tracker lives in `dd._aux[:fuse29]`.
- `build_fuse29_actuators` (29 channels, raw units, at a requested time, from `:zmq` aux records or `:dd` fields) replaces `build_fpe_sequences_from_aux`.
- New parameters: `nn_hmode_enter` (0.7), `nn_hmode_exit` (0.3), `nn_warmup_time` (replay actuators from that time before the first prediction; the model needs ~1 s to settle from a zero state).
- ne / Te / Ti are pinned at ρ_tor = 0.85 (the model's radius) rather than FUSE's 0.9 convention; the four H-mode-only heads are NaN-gated diagnostics in `nn_prediction`. Te/Ti are applied in L-mode too via `blend_core_edge_Lmode` when `nn_ped_quantities = :all`.

**Tests / data / docs**
- `test/test_pedestal_nn_predictor.jl`: contract, manifest/auto-fetch surface, replay of the shipped reference shot 199055 through both graphs (must match the model authors' outputs to 1e-3), functional-step and tracker semantics — 90 tests.
- `test/smoke_pedestal_nn_shot.jl`: synthetic beam ramp through the ZMQ actuator path with an L→H transition — 71 tests.
- `data/pedestal_nn`: `history_morning.npz` removed, `fuse29_reference_shot.npz` (24 KB) added, README rewritten. Deploy files: comment-only size updates; `FUSE_PEDESTAL_NN_DIR` semantics unchanged.

No `Project.toml` changes.

## Verification

- [x] `julia --project test/test_pedestal_nn_predictor.jl` — 90/90 pass, reference replay to 1e-3 on both graphs
- [x] `julia --project test/smoke_pedestal_nn_shot.jl` — 71/71 pass
- [x] `using FUSE` precompiles
- [ ] Short `ActorDynamicPlasma` run with `ne_from=:nn_predictor`, `fpe_source=:dd`, δt = 0.05 s (`experiment_predictive_rt` style) — not yet run

## Caveats worth knowing (measured on the model itself)

- Under constant actuators the continuous heads settle within ~5 s, but the `hmode` probability decays slowly beyond the ~12.8 s training window (the state encodes elapsed shot time). The hysteretic gate absorbs this; treat the gate with suspicion in very long steady-state runs.
- `ech_total` carries an inherited scale factor from the upstream DIII-D pipeline (training p99 ≈ 30 "MW"); pass nominal MW.

https://claude.ai/code/session_01HG6NsgJwK86Hj7WzNSUHAf


nanshi1177:
Pushed two fixes to this branch. The pinj/ech_total point-sampling one is the substantive one — DIII-D beam chopping is phase-locked to the 50 ms tick, so shot 200000 reads 0 MW at every tick from 2.5 to 4.0 s while the true windowed mean is 0.95 MW. Verified against pulse_schedule.nbi directly. The 90 tests in test_pedestal_nn_predictor.jl still pass.

PR #1180, opened 2026-09-18
Coupling Robustness Changes

prechelg:
1) progress_ActorDynamicPlasma: display-only try/catch — get_from(:ip) could throw on a transiently inconsistent dd right after an equilibrium-slice restore and kill the run from the progress meter
2) ActorZMQ receive!: clamp wire NBI power to [0,5e6] W and voltage to [0,2e5] eV 
3) ActorZMQ receive!: guard the first-step full flux_surfaces (snapshot at psizr commit, restore + retry next exchange on failure since marginal handoff maps can yield boundary.psi = nothing mid-write)
4) ActorFluxMatcher._step: floor Te/ne/Ti/ni at entry because numerically negative-zero profile values (-2e-11) blew up x^y in collision_frequencies
5) ActorEquilibrium restores: empty! before IMAS.fill!


nanshi1177 (approved):

nanshi1177:
Added accounting on top of your guards, the guards themselves are unchanged. A skipped substep advances time without applying its operator, and maxlog hid repeats, so a run could skip physics for seconds and still report ok.

Track failures per substep (total and consecutive), abort on a streak via max_consecutive_substep_failures, and print the tally at the end. Left unset the limit is 10 with ActorZMQ enabled and 1 otherwise, since the guards also affect standalone studies that relied on failing loudly.

ActorFluxMatcher counts its soft-fail paths too; they never throw so they never reach the guard. Skip the snapshot when a substep is disabled.

prechelg:
> Added accounting on top of your guards, the guards themselves are unchanged. A skipped substep advances time without applying its operator, and maxlog hid repeats, so a run could skip physics for seconds and still report ok.
> 
> Track failures per substep (total and consecutive), abort on a streak via max_consecutive_substep_failures, and print the tally at the end. Left unset the limit is 10 with ActorZMQ enabled and 1 otherwise, since the guards also affect standalone studies that relied on failing loudly.
> 
> ActorFluxMatcher counts its soft-fail paths too; they never throw so they never reach the guard. Skip the snapshot when a substep is disabled.

Looks good, thank you Nan.

nanshi1177 (approved):

